PR #94, opened 2014-11-30
obf-generation/regions/indexes.xml modified to add Spain regions.

helfio:
I'm not sure if I've modified correctly the file. Please check and tell me if I have to change anything. Thanks.


vshcherb:
There is no spain regions on geofabrik website "http://download.geofabrik.de/europe/spain{0}.osm.pbf
(http://www.geofabrik.de/)

PR #97, opened 2015-01-08
Update regions; Add split Netherlands as inactivated set

hvdwolf:
The Netherlands split has been added but "out-commented" to make it possible to activate it at the right moment


vshcherb:
The osm-downloads is missing Netherlands itself http://download.osmand.net/osm-downloads/, So regions split will not be active now.

PR #99, opened 2015-02-12
New server for some data files

hbogner:
Added our data server (http://data.osm-hr.org/osm/) to indexes.xml and commented those countries on your servers.
These exports have optimized borders, and are refreshed every 6 hours.


xmd5a2:
We do not want to download more data from external sources.

hbogner:
OK.

PR #107, opened 2016-04-10
[OsmAndMapCreator] Add buttons to select Start and End points using screen center

rominf:
Reason: to select start and finish points quickly and precisely using these buttons and "Select point..." dialog.


vshcherb:
Good intention but we do need to make it optional for developer mode.

rominf:
@vshcherb is it OK now?

PR #126, opened 2016-06-09
MapCreator: process boundary=postal_code, write postcode for buildings

rominf:


OsmAndMapCreator/src/net/osmand/data/Multipolygon.java
```diff
@@ -161,6 +182,10 @@ public LatLon getCenterPoint() {
 		return OsmMapUtils.getWeightCenterForNodes(points);
 	}
 
+	public void intersectsWith(Multipolygon multipolygon) {
```
vshcherb: Empty method?

PR #143, opened 2017-06-14
Missing motorway_links breaks routing badly

elgaard:
Osmand routing fails with missing mororway_link's. Or even produces very bad routes.

### DIFF
--- a/obf-generation/basemap/query_roads_osm.py
+++ b/obf-generation/basemap/query_roads_osm.py
@@ -76,6 +76,7 @@ def process_roads(cond, filename, fields):
 
 if __name__ == "__main__":
 	process_roads("highway='motorway'", "line_motorway.osm", ['highway', 'junction', 'route'])
+        process_roads("highway='motorway_link'", "line_motorway.osm", ['highway', 'junction', 'route'])
 	process_roads("highway='trunk'", "line_trunk.osm", ['highway', 'junction', 'route'])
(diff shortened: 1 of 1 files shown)


xmd5a2:
This script is used for basemap generation. No routing is possible via basemap. Can you explain issue more clearly?

elgaard:
Osmand (v 2.6.3) does route using the basemap, when no other maps are available.
Maybe this should be disabled, because it is worse than nothing. Especially when there are middle maps for the start and destination but not in between.

But even just looking at the basemap, it cannot be right that there are holes in the motorway network. E.g, look at
http://www.openstreetmap.org/#map=15/47.7735/12.9811

in the basemap.

xmd5a2:
You must have been mistaken. We don't have routing via basemap. May be you are using another routing provider? Try deleting all maps except basemap and calculate a route.
Anyway, motorway_link IS used by our routing engine https://github.com/osmandapp/OsmAnd-resources/blob/master/routing/routing.xml#L228
Holes in motorway networks at basemap is a separate issue and is not related to routing.

elgaard:
Sorry, yes I was wrong. Osmand does try to route, even when there is no map. But it does not really use the basemap. It just produces routes that are nonsense:
neubrandenburg

Anyway, I still find it confusing that motorways are not connected in the basemap:

salzburg

PR #179, opened 2017-10-02
Improved OSM Live map generation

PaulStets:


OsmAndMapCreator/src/net/osmand/data/diff/ObfDiffGenerator.java
```diff
@@ -206,6 +244,17 @@ private void compareMapData(ObfFileInMemory fStart, ObfFileInMemory fEnd, boolea
...
 	
+	private EntityId getMapEntityId(long id) {
+		if (id < (1 << 41)) {
+			if ((id % 2) == 0) {
+				return new EntityId(EntityType.WAY, id >> (BinaryInspector.SHIFT_ID + 1));
+			} else {
+				return new EntityId(EntityType.NODE, id >> (BinaryInspector.SHIFT_ID + 1));
+			}
+		}
+		return new EntityId(EntityType.RELATION, id >> (BinaryInspector.SHIFT_ID + 1));
```
vshcherb: should be return null as it is impossible to detect

OsmAndMapCreator/src/net/osmand/data/diff/ObfDiffGenerator.java
```diff
@@ -188,10 +215,21 @@ private void compareMapData(ObfFileInMemory fStart, ObfFileInMemory fEnd, boolea
 					}
 				} else {
 					if (objE == null) {
-						// Object with this id is not present in the second obf
-						BinaryMapDataObject obj = new BinaryMapDataObject(idx, objS.getCoordinates(), null,
-								objS.getObjectType(), objS.isArea(), new int[] { deleteId }, null);
-						endData.put(idx, obj);
+						if (deletedObjIds != null) {
+							if (deletedObjIds.containsKey(osmid)) {
```
vshcherb: Not needed check? why storing long & checking long if we need to check EntityId? WTF?

PR #267, opened 2018-07-16
New api

vshcherb:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/ApiController.java
```diff
@@ -0,0 +1,29 @@
...
+
+    private static final Log LOGGER = LogFactory.getLog(ApiController.class);
+
+    @Autowired
+    UpdateIndexes updateIndexes;
+
+    @RequestMapping(value = "update", method = RequestMethod.GET)
+    public String update(Model model) throws IOException {
+        updateIndexes.update();
+        return "pub/indexes";
```
vshcherb: We should keep original urls!

java-tools/OsmAndServer/src/main/java/net/osmand/server/index/dao/AbstractDAO.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class AbstractDAO {
+
+    private static final Path BASE_PATH = Paths.get("/var/www-download/");
```
vshcherb: This definitely should be parameterizable and better to use application.properties

java-tools/OsmAndServer/src/main/java/net/osmand/server/index/reader/FileDirectoryReader.java
```diff
@@ -0,0 +1,26 @@
+package net.osmand.server.index.reader;
+
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
```
vshcherb: We have already utiltities in Algorithms.

PR #273, opened 2018-07-23
New api

vshcherb:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/IndexController.java
```diff
@@ -41,17 +62,75 @@ public FileSystemResource indexesPhp(@RequestParam(defaultValue="", required=fal
...
+                          Model model) throws IOException {
+        File fl = null;
+        if (sortby == null || sortby.isEmpty()) {
+            // Update at first load
+            fl = downloadIndexes.getIndexesXml(true, false);
+        } else {
+            // Do not update when filtering
+            fl = downloadIndexes.getIndexesXml(false, false);
+        }
+        DownloadIndexDocument doc = unmarshallIndexes(fl);
```
vshcherb: Too much copy paste, almost every line regions.stream().sorted(Comparator.comparing(DownloadIndex::getDate).... Must be a generic function for that?

java-tools/OsmAndServer/src/main/java/net/osmand/server/index/DownloadIndex.java
```diff
@@ -1,136 +1,130 @@
...
-    private long contentSize = -1;
-    private String name;
-    private String description = null;
+    private long timestamp;
+
+	@XmlAttribute(name = "date")
+    private String date;
+
+    @XmlJavaTypeAdapter(DownloadIndexSizeAdapter.class)
+	@XmlAttribute(name = "size")
```
vshcherb: Again something with spaces / tabs in the same commit!
+    @XmlJavaTypeAdapter(DownloadIndexSizeAdapter.class)
+	@XmlAttribute(name = "size")

PR #287, opened 2018-08-21
host fix

Evgenij27:


java-tools/OsmAndServer/src/main/java/net/osmand/server/services/images/ImageService.java
```diff
@@ -277,8 +281,9 @@ public CameraPlace processWikimediaData(double lat, double lon, String osmImage)
     }
 
     private static class OsmandPhotoApiConstants {
-        static final String OSMAND_GET_IMAGE_ROOT_URL = "https://osmand.net/api/mapillary/get_photo";
-        static final String OSMAND_PHOTO_VIEWER_ROOT_URL = "https://osmand.net/api/mapillary/photo-viewer";
+        static final String OSMAND_SCHEME = "https";
```
vshcherb: This should be also read from original request

PR #288, opened 2018-08-21
Initial commit

Evgenij27:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/ApiController.java
```diff
@@ -37,10 +41,12 @@
     private String procFile;
 
     private final ImageService imageService;
```
vshcherb: It is better to used autowired and not to put all services in controller.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/ApiController.java
```diff
@@ -140,4 +146,21 @@ public String getPhotoViewer(@RequestParam("photo_id") String photoId, Model mod
...
+                                   @RequestParam(required = false) Integer nd,
+                                   @RequestParam(required = false) Integer ns,
+                                   @RequestParam(required = false) String lang,
+                                   @RequestParam(required = false) String os,
+                                   @RequestParam(required = false) String aid,
+                                   @RequestParam(required = false) String discount) throws IOException {
+        boolean iosVersion = os != null && os.equals("ios");
+        boolean appVersion3 = version != null && version.startsWith("3.");
+
+        System.out.println(System.getProperty("user.dir"));
```
vshcherb: System out is not good to merge

java-tools/OsmAndServer/src/main/resources/application.yml
```diff
@@ -56,10 +56,12 @@ mapillary:
 osmlive:
   status: /var/www-download/api/.proc_timestamp
 
-
 monitoring:
   enabled: ${ENABLE_MONITORING:false}
 
+motd:
```
vshcherb: settings is not a parameter it self and should use standard location /var/www-download/ for all settings, configs etc. So we are not going to pollute yml with 10-20 same locations

java-tools/OsmAndServer/src/main/resources/application.yml
```diff
@@ -73,6 +75,14 @@ download:
   files: /var/www-download/
   load: 60
 
+mapillary:
+  clientid: LXJVNHlDOGdMSVgxZG5mVzlHQ3ZqQTo0NjE5OWRiN2EzNTFkNDg4
+
+osmlive:
+  status: /var/www-download/api/.proc_timestamp
+
+---
```
vshcherb: This line duplicate is not clear at all

vshcherb:
The request is far from being ready, lets reiterate it later

PR #292, opened 2018-08-23
exclude imgs with 0 ca

Evgenij27:


java-tools/OsmAndServer/src/main/java/net/osmand/server/services/images/ImageService.java
```diff
@@ -128,9 +128,9 @@ private boolean angleDiff(double angle, double diff) {
     private void splitCameraPlaceByAngel(CameraPlace cp, Map<String, List<CameraPlace>> resultMap) {
         double ca = cp.getCa();
         double bearing = cp.getBearing();
-        if (ca >= 0 && angleDiff(bearing - ca, 30.0)) {
+        if (ca > 0d && angleDiff(bearing - ca, 30.0)) {
             resultMap.get(RESULT_MAP_ARR).add(cp);
-        } else if (ca >= 0 && angleDiff(bearing - ca, 60.0) || ca < 0) {
+        } else if (ca > 0d && angleDiff(bearing - ca, 60.0) || ca < 0d) {
```
vshcherb: here should be 
"} else if ((ca > 0d && angleDiff(bearing - ca, 60.0)) || !(ca > 0d)) {"
probably ca <= 0d but that should be tested

Otherwise all 0d will be lost which is not the logic and it should go to RESULT_MAP_HALFVISARR

PR #299, opened 2018-09-13
add check live

Evgenij27:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/IndexController.java
```diff
@@ -138,4 +205,49 @@ public String listPhp(@RequestParam(required = false) String sortby,
...
+                    if (filename.startsWith(file)) {
+                        writeAttributes(xmlWriter, mapFile, timestamp);
+                    }
+                }
+            }
+            xmlWriter.writeEndElement();
+            xmlWriter.writeEndDocument();
+            xmlWriter.flush();
+        } catch (XMLStreamException ex) {
+            LOGGER.error(ex.getMessage(), ex);
```
vshcherb: no need for stack trace - the log is huge with it. LOGGER.error(" Error within checklive: "+ ex.getMessage());

PR #307, opened 2018-10-19
register/delete device

Evgenij27:


java-tools/OsmAndServer/src/main/java/net/osmand/server/assist/data/NewDevice.java
```diff
@@ -0,0 +1,49 @@
+package net.osmand.server.assist.data;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.telegram.telegrambots.api.objects.User;
+
+public class NewDevice {
```
vshcherb: We already have DeviceBean why would we need new class which mostly isdpulciate

PR #318, opened 2019-04-24
updated windows batch file to inlcude option with increased heap memory

pyramid3d:
Converting very large pbf files (tested with Thailand) leads to an "out of java heap memory error" on windows when running with the default parameters.

To have an alternative command line option embedded in the batch file helps a lot and avoids unnecessary look-ups to find the right parameters.


pyramid3d:
Thanks for merging! Cheerios.

PR #342, opened 2020-02-09
Map download script in python 

ppenguin:
Map download script in python for air-gapped devices, to download wanted countries (based on a wanted list) to e.g. an SD card of e.g. a head-unit.

It's a quick hack but functional, might be helpful. It should be trivial to make it more robust and to add sub-areas (counties, cities) to it (framework already there).

The only afterthought is that it could make the download limit for the free version obsolete, so maintainer might want to consider whether it should be publicly available, or whether there is any way to still motivate people to buy based on the limitation?

(Of course I myself am using the paid version OsmAnd+)


vshcherb:
There is no plan to support that script on our side.
The question, whether it should be open or not, is discussible obviously there is no guarantee that this API will stay open if it is going to be used by majority of users.

ppenguin:
Ok, fair enough, just thought to contribute it in case somebody finds it useful, but I can see the potential side-effects...
Cheers

PR #357, opened 2020-06-30
Fix Transport Index issue (should be accepted once release is out)

MadWasp79:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/BinaryMapIndexWriter.java
```diff
@@ -1397,10 +1397,11 @@ public void writeTransportStop(long id, int x24, int y24, String name, String na
 	public void writeIncompleteTransportRoutes(Collection<net.osmand.data.TransportRoute> incompleteRoutes, Map<String, Integer> stringTable) throws IOException {
 		checkPeekState(TRANSPORT_INDEX_INIT);
 		OsmandOdb.IncompleteTransportRoutes.Builder irs = OsmandOdb.IncompleteTransportRoutes.newBuilder();
+		long fp = getFilePointer() + 3;
```
vshcherb, line 1400: 3... not constant

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/BinaryMapIndexWriter.java
```diff
@@ -1155,11 +1155,14 @@ private int registerString(Map<String, Integer> stringTable, String s) {
 		return size;
 	}
 
+	private long transportIndexOffset = 0;
```
vshcherb, line 1158: No fields, local variable is enough inside IndexTransportCreator

PR #369, opened 2020-11-09
#9693 (lcn=yes) fix

MadWasp79:
In case there is no relation, but are still network tags on segment - we add tags to correctly mark segments with bicycle routes 'on'.


vshcherb:
Looks weird is it possible with rendering_types entity convert to do the same?

PR #382, opened 2021-02-12
Add diff between 26 and 24 zoom for precision results

ivanPyrohivskyi:
Diff between 26 and 24 zooms for precision results


vshcherb:
Should be accepted after https://github.com/osmandapp/OsmAnd/pull/10848

PR #385, opened 2021-02-25
Gradient colorization for route

ivanPyrohivskyi:
- [ ] Require RouteColorize from https://github.com/osmandapp/OsmAnd/pull/11009

- [ ] For check in the Map Creator: "Load GPX file ... ", after will be able "Colorize GPX file" and use one from variants: "Altitude", "Speed", "Slope"


ivanPyrohivskyi:
Added possibility of load different color schemes: decimal RGB, RGB, RGBA

PR #388, opened 2021-03-15
Set city polygon type according admin_centre

ivanPyrohivskyi:
Set CityType according admin_centre type (tag "place")

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexAddressCreator.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexAddressCreator.java
@@ -373,6 +373,13 @@ public class IndexAddressCreator extends AbstractIndexPartCreator {
 				ct = CityType.SUBURB;
 			}
 		}
+		if (ct == null && e instanceof Relation) {
+			ctx.loadEntityRelation((Relation) e);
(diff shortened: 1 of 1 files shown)


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexAddressCreator.java
```diff
@@ -373,6 +373,13 @@ private Boundary extractBoundary(Entity e, OsmDbAccessorContext ctx) throws SQLE
 				ct = CityType.SUBURB;
 			}
 		}
+		if (ct == null && e instanceof Relation) {
+			ctx.loadEntityRelation((Relation) e);
```
vshcherb, line 377: before loading relation, it  could be checked if admin_centre exist at all.

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexAddressCreator.java
```diff
@@ -373,6 +373,13 @@ private Boundary extractBoundary(Entity e, OsmDbAccessorContext ctx) throws SQLE
...
+			List<Entity> adminCentres = ((Relation) e).getMemberEntities("admin_centre");
```
vshcherb, line 378: Changes are not clear cause it doesn't look correct for https://www.openstreetmap.org/relation/1682189#map=13/47.3697/8.7966

PR #406, opened 2021-06-14
Add route_radius tag

Dima-1:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/OsmGpxWriteContext.java
```diff
@@ -107,6 +107,7 @@ public void writeTrack(OsmGpxFile gpxInfo, Map<String, String> extraTrackTags, G
 				serializer.attribute(null, "lon", latLonFormat.format(gpxFile.findPointToShow().lon));
 				tagValue(serializer, "route", "segment");
 				tagValue(serializer, "route_type", "track");
+				tagValue(serializer, "route_radius", gpxFile.getOuterRadius());
```
vshcherb, line 110: we should have limited number of values: 5, 10, 50, 100, 500, 1000, 5000, 10000

vshcherb: A - 5, B - 10, C - 50, D - 100,  E - 500, F - 1000, G - 5000, H -  10000

( (int) (10 ^ ((X - 'A' + 1) / 2) * 5 + 1) / 5)

vshcherb: int d = x - 'a' + 1;
int r = (int) Math.pow(10, ((x - 'a' + 1) / 2)) * (d % 2 == 0 ? 1 : 5);
System.out.println(x + " " + r);

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/server/osmgpx/DownloadOsmGPX.java
```diff
@@ -136,7 +129,7 @@ public static void main(String[] args) throws Exception {
 				switch (s[0]) {
 				case "--acitivity-type":
 					if (val.trim().length() > 0) {
-						qp.activityTypes = new TreeSet<RouteActivityType>();
```
vshcherb, line 139: why it's updated ?

Dima-1: Because Exception in thread "main" java.lang.ClassCastException: class net.osmand.osm.RouteActivityType cannot be cast to class java.lang.Comparable

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/server/osmgpx/DownloadOsmGPX.java
```diff
@@ -537,19 +528,12 @@ protected void downloadGPXMain() throws Exception {
...
-					for (WptPt p : s.points) {
-						r.minlat = Math.min(r.minlat, p.lat);
-						r.maxlat = Math.max(r.maxlat, p.lat);
-						r.minlon = Math.min(r.minlon, p.lon);
-						r.maxlon = Math.max(r.maxlon, p.lon);
-					}
-				}
-			}
+		if (gpxFile.error == null) {
+			QuadRect rect = gpxFile.getRect(r.lon, r.lat, r.lon, r.lat);
```
vshcherb, line 532: this is problematic in original request as well

PR #407, opened 2021-06-16
add process tag turn:lanes:both_ways

alisa911:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/osm/MapRoutingTypes.java
```diff
@@ -176,6 +176,7 @@ private boolean testNonParseableRules(String tag, String value) {
 
 	public boolean encodeEntity(Map<String, String> tags, TIntArrayList outTypes, Map<MapRouteType, String> names){
 		boolean init = false;
+		processBothWaysTag(tags);
```
vshcherb, line 179: his method should be somewhere in transformTags

PR #421, opened 2021-08-09
Fix #12297, close osmandapp/OsmAnd#12324 missing column in the travel articles

Dima-1:


Dima-1:
osmandapp/OsmAnd#12297
osmandapp/OsmAnd#12324

PR #427, opened 2021-09-01
Fix incorrect bus stop name

Dima-1:


Dima-1:
Fix osmandapp/OsmAnd#12292

PR #443, opened 2021-11-17
move night builds to dl2, fix BuildsController

Dima-1:


java-tools/OsmAndServer/src/main/resources/application.yml
```diff
@@ -84,6 +84,7 @@ geoip.url: "http://localhost:8081/json/"
 ---
 spring.jpa.hibernate.ddl-auto: validate
 
+night-builds-json.location: dl2.osmand.net/night-builds/night-builds-json/
```
vshcherb, line 87: delete this place as wrong location

PR #451, opened 2021-12-17
Visibility compare

Dima-1:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/IconVisibilityComparator.java
```diff
@@ -0,0 +1,176 @@
...
+import org.xmlpull.v1.XmlPullParser;
+import org.xmlpull.v1.XmlPullParserException;
+
+import java.io.*;
+import java.util.*;
+
+import static net.osmand.binary.BinaryMapIndexReader.*;
+
+public class IconVisibilityComparator {
+	String RENDER_FILE = "/home/user/osmand/issues/i985/default.render.xml";
```
vshcherb, line 18: Path should be relative and it should be using default render xml and synthtetic. Through parameters or env it could be overridden

vshcherb:
We need full recreate the branch and delete Syntetic.obf from the test because it's already present

vshcherb:
Commit - Remove default.render.xml doesn't delete file from the branch!

PR #452, opened 2021-12-30
Heightmap support

alexey-pelykh:


cpp-tools/map-viewer/main.cpp
```diff
@@ -791,10 +800,14 @@ void keyboardHandler(unsigned char key, int x, int y)
         }
         else
         {
-            if (wasHeightsDirSpecified)
+            if (heightsCollection)
             {
-                //auto provider = new OsmAnd::HeightmapTileProvider(heightsDir, cacheDir.absoluteFilePath(OsmAnd::HeightmapTileProvider::defaultIndexFilename));
-                //renderer->setElevationDataProvider(std::shared_ptr<OsmAnd::IMapElevationDataProvider>(provider));
+                renderer->setElevationDataProvider(
+                    std::make_shared<OsmAnd::SqliteHeightmapTileProvider>(heightsCollection, 33)
```
StarQuasar, line 806: What does it mean `33`?
May be there are make sense to put it at the top of this file as a const/constexpr value?
If this value can be configured at all.

alexey-pelykh: Valid point, it should be part of renderer interface

PR #453, opened 2021-12-31
Visibility compare

Dima-1:


java-tools/OsmAndMapCreator/src/test/java/net/osmand/obf/IconVisibilityTest.java
```diff
@@ -0,0 +1,268 @@
...
+								}
+
+								@Override
+								public boolean isCancelled() {
+									return false;
+								}
+							});
+
+					reader.searchMapIndex(req, mapIndex);
+					System.out.printf("zoom %d total objects: %d ", zoom, statCounts[ALL]);
```
vshcherb, line 133: sysout not needed - will be on every call

PR #474, opened 2022-04-13
add delete osm_recipients

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailRegistryService.java
```diff
@@ -6,18 +6,30 @@
...
+	private final JdbcTemplate jdbcTemplate;
+	
+	private final OsmRecipientsRepository recipientsRepository;
+	
+	public EmailRegistryService(OsmRecipientsRepository recipientsRepository, JdbcTemplate jdbcTemplate) {
+		this.recipientsRepository = recipientsRepository;
+		this.jdbcTemplate = jdbcTemplate;
+	}
+	
+	@Transactional
```
vshcherb, line 28: here transactional not needed - check unsubscribedRepo.deleteAllByEmail(email);

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailRegistryService.java
```diff
@@ -6,18 +6,30 @@
...
+import org.springframework.transaction.annotation.Transactional;
 
 @Service
 public class EmailRegistryService {
-
-	@Autowired
-	private JdbcTemplate jdbcTemplate;
-
+	
+	private final JdbcTemplate jdbcTemplate;
```
vshcherb, line 19: Doesn't follow conventions we use everywhere  Autowired so here it's not clear why should we do different

alisa911: @Autowired isn't recommended. On the official page Spring, they recommend to use Constructor-based Dependency Injection. @Autowired has many restrictions in use, hides interface and other. Intelij Idea official highlights @Autowired us unrecommended.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailRegistryService.java
alisa911: I added it after exception No EntityManager. I read it need for deleteBy custom method

PR #475, opened 2022-04-18
Fix names IndexOutOfBoundsException

Dima-1:


vshcherb:
We want to limit dictionary to 8000 (not to extend to 64K)

PR #491, opened 2022-06-16
add filter map

alisa911:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/NativeJavaRendering.java
```diff
@@ -482,60 +485,35 @@ public Map<String, FileIndex> initIndexesCache(File dir,
...
-			while (fileIt.hasNext()) {
-				File obf = fileIt.next();
-				try {
-					FileIndex fileIndex = cache.getFileIndex(obf, true);
-					if (fileIndex != null) {
-						map.put(obf.getAbsolutePath(), fileIndex);
-					}
-					boolean basemapPresent = checkMoreGenericMapPresent(allFileNames, obf.getName());
-					if (filterDuplicates && basemapPresent) {
-						log.debug("SKIP initializing cause bigger map is present: " + obf.getName());
```
vshcherb, line 505: important log message was deleted 						log.debug("SKIP initializing cause bigger map is present: " + obf.getName());

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/NativeJavaRendering.java
```diff
@@ -482,60 +485,35 @@ public Map<String, FileIndex> initIndexesCache(File dir,
...
-						fileIt.remove();
-					}
-				} catch (Exception e1) {
-					e1.printStackTrace();
+			
+			List<String> regionNameList = new ArrayList<>();
+			List<File> files = new ArrayList<>();
+			
+			OsmandRegions osmandRegions = new OsmandRegions();
```
vshcherb, line 501: This code is now added in all variations where jave rendering is initialized in all cases but it's needed likely only for filterDuplicates

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/util/FilterMap.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.List;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+
+public class FilterMap {
+    
+    FilterMap() {
+    }
+    
+    public static boolean checkBiggerMapExist(List<BinaryMapIndexReader> files, List<String> regionNameList, String fileName, OsmandRegions osmandRegions) {
```
vshcherb, line 17: 1. all static methods doesn't look good
2. parameter list is rather long
3. some methods should be private
4. Naming: FilterMap? Suggestion: UtilityToExcludeDuplicatedMaps

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/NativeJavaRendering.java
alisa911: +

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/util/FilterMap.java
alisa911: 1. `+`
2. partially
3. `+`
4. `+`

PR #506, opened 2022-09-05
Track info

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/GpxController.java
```diff
@@ -304,6 +305,69 @@ public ResponseEntity<String> getGpxInfo(@RequestPart(name = "file") @Valid @Not
 		}
 	}
     
+    @PostMapping(path = {"/get-track-data"}, produces = "application/json")
```
vshcherb, line 308: Post != get-track-data (get wrong word)

Probably should be "process-track-data", "retrieve-?"

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/GpxController.java
```diff
@@ -304,6 +305,69 @@ public ResponseEntity<String> getGpxInfo(@RequestPart(name = "file") @Valid @Not
...
+    @ResponseBody
+    public ResponseEntity<String> getGpxInfo(@RequestPart(name = "file") @Valid @NotNull @NotEmpty MultipartFile file,
+                                             HttpSession httpSession) throws IOException {
+        
+        File tmpGpx = File.createTempFile("gpx_" + httpSession.getId(), ".gpx");
+        InputStream is = file.getInputStream();
+        FileOutputStream fous = new FileOutputStream(tmpGpx);
+        Algorithms.streamCopy(is, fous);
+        is.close();
+        fous.close();
```
vshcherb, line 318: FIle never deleted?

PR #512, opened 2022-09-14
add deep copy

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/GpxController.java
```diff
@@ -319,7 +320,7 @@ public ResponseEntity<String> processTrackData(@RequestPart(name = "file") @Vali
 			WebGpxParser.TrackData gpxData = new WebGpxParser.TrackData();
 			
 			GPXTrackAnalysis analysis = getAnalysis(gpxFile, false);
-			GPXTrackAnalysis srtmAnalysis = getAnalysis(gpxFile, true);
+			GPXTrackAnalysis srtmAnalysis = getAnalysis(SerializationUtils.clone(gpxFile), true);
```
vshcherb, line 323: could be attached after 	webGpxParser.addRoutePoints(gpxFile, gpxData);

PR #515, opened 2022-09-19
add delete point gpx

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/GpxController.java
```diff
@@ -341,6 +346,18 @@ public ResponseEntity<String> processTrackData(@RequestPart(name = "file") @Vali
 		}
 	}
 	
+	@PostMapping(path = {"/get-track-points-between-two-route-points"}, produces = "application/json")
```
vshcherb, line 349: url -> "/build-route", needs 2 points and probably should be in RoutingContronller

PR #523, opened 2022-10-11
Speed up osm generation especially for osmand live

vshcherb:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/osm/MapRenderingTypesEncoder.java
```diff
@@ -820,8 +832,7 @@ protected List<EntityConvert> getApplicableConverts(Map<String, String> tags, En
 								break;
 							}
 						}
-						if (checkConvert(tags, ec, entity) && ec.type == filter &&
-								ec.applyToType.contains(appFilter)) {
+						if (checkConvert(tags, ec, entity)) {
```
ivanPyrohivskyi, line 835: Is save to delete ec.type == filter && ec.applyToType.contains(appFilter)

PR #526, opened 2022-10-20
Add center node for island with coastline

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/15483

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexVectorMapCreator.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexVectorMapCreator.java
@@ -736,6 +736,21 @@ public class IndexVectorMapCreator extends AbstractIndexPartCreator {
 					}
 				}
 			}
+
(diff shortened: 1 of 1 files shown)


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexVectorMapCreator.java
```diff
@@ -736,6 +736,21 @@ public void iterateMainEntity(Entity e, OsmDbAccessorContext ctx, IndexCreationC
 					}
 				}
 			}
+
+			//create center node for island/islet with coastline
+			if (e instanceof Way && tags.size() > 2 && "coastline".equals(tags.get("natural"))
+					&& ("island".equals(tags.get("place")) || "islet".equals(tags.get("place")))) {
+				QuadRect bbox = ((Way) e).getLatLonBBox();
+				Node node = new Node(bbox.centerY(), bbox.centerX(), e.getId());
```
vshcherb, line 744: id looks weird could be in conflict with other random point on the map

PR #554, opened 2022-12-15
Web server

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -225,6 +226,28 @@ public ResponseEntity<String> deleteFile(@RequestParam String name, @RequestPara
...
+		if (dev == null) {
+			return userdataService.tokenNotValid();
+		} else {
+			if (updatetime != null) {
+				fl = filesRepository.findTopByUseridAndNameAndTypeAndUpdatetime(dev.userid, name, type,
+						new Date(updatetime));
+			}
+			if (fl == null) {
+				return userdataService.error(UserdataService.ERROR_CODE_FILE_NOT_AVAILABLE, "File is not available");
+			}
```
vshcherb, line 244: This code is copy-paste and it should be generalized inside service

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -225,22 +226,35 @@ public ResponseEntity<String> deleteFile(@RequestParam String name, @RequestPara
...
-	                                         @RequestParam String name, @RequestParam String type) throws IOException {
+	                                         @RequestParam String name,
+	                                         @RequestParam String type) throws IOException {
 		PremiumUserDevice dev = checkUser();
 		if (dev == null) {
 			return userdataService.tokenNotValid();
 		}
-		PremiumUsersRepository.PremiumUser pu = usersRepository.findById(dev.userid);
+		PremiumUsersRepository.PremiumUser pu = userdataService.getUserById(dev.userid);
 		ResponseEntity<String> validateError = userdataService.validateUser(pu);
```
vshcherb, line 252: don't expose getuser by id make method userdataService.validateUser(dev.userid)

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -445,4 +428,58 @@ public ResponseEntity<String> checkDownload(@RequestParam(value = "file_name", r
...
+	                                        @RequestParam String updatetime) throws IOException {
+		WebGpxParser.Wpt wpt = gson.fromJson(data, WebGpxParser.Wpt.class);
+		
+		PremiumUserDevice dev = checkUser();
+		ResponseEntity<String> validateError = userdataService.validate(dev);
+		if (validateError != null) {
+			return validateError;
+		}
+		
+		return userdataService.addOrDeleteFavorite(wpt, fileName, dev, fileType, updatetime, DELETE_FAVORITE);
```
vshcherb, line 446:
```
addOrDeleteFavorite - could be part of 
```

PR #610, opened 2023-04-24
add promo

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/PromoService.java
```diff
@@ -0,0 +1,125 @@
...
+            PremiumUsersRepository.PremiumUser existingUser = usersRepository.findByEmail(email);
+            if (existingUser != null && !resp.error) {
+                promo.used++;
+                promoRepository.save(promo);
+                return ResponseEntity.ok(expireTime.toString());
+            } else {
+                return ResponseEntity.badRequest().body(resp.deviceSub.purchaseToken);
+            }
+        }
+        return ResponseEntity.badRequest().body("Limit promo is exceeded!");
```
vshcherb, line 53: - 'Unfortunately we ran out of available promocodes'. 
- Also email should be sent. Request shouldn't be Error or should be handled correctly on UI to display the message.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/PromoService.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+    
+    public ResponseEntity<String> addUser(String name, String email) {
+        PromoRepository.Promo promo = promoRepository.findByName(name);
+        if (promo.used < promo.numberLimit) {
+            String key = "promo_" + promo.name;
+            Date expireTime = getExpirationDate(promo);
+            PromoResponse resp = createPromoSubscription(email, key, expireTime);
+            PremiumUsersRepository.PremiumUser existingUser = usersRepository.findByEmail(email);
```
vshcherb, line 44: if existing user already exists it doesn't make any sense to register it again, unless it was an expired subscription before

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/PromoController.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.web.bind.annotation.RequestParam;
+
+@RequestMapping("/promo")
+@Controller
+public class PromoController {
+    
+    @Autowired
+    PromoService promoService;
+    
+    @PostMapping(path = {"/add-user"})
```
vshcherb, line 18: 1 method could reuse ApiController

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/PromoService.java
```diff
@@ -0,0 +1,125 @@
...
+            if (existingUser != null && !resp.error) {
+                promo.used++;
```
vshcherb, line 46: What if users will be registered concurrently ? should be the comment or the proper fix that field used could be bigger than limit?

vshcherb:
I don't see any schema changes which I think should be in this pull request

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/PromoService.java
alisa911: Messages with error responses were added in UI (resp.deviceSub.purchaseToken from createPromoSubscription can have: new PRO subscription is updated, user already has PRO subscription or please enter email only if email isn't correct.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/PromoService.java
alisa911: this is implemented in createPromoSubscription

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/PromoController.java
alisa911: > I don't see any schema changes which I think should be in this pull request

PromoRepository

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/AdminController.java
```diff
@@ -492,7 +436,22 @@ public String index(Model model) throws SQLException {
 		return "admin/info";
 	}
 	
-
+	private Map<String, String> getLastUsersPromo(List<PromoRepository.Promo> promos) {
```
vshcherb, line 439: Method is too slow, will be an issue once it will be growing

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/AdminController.java
```diff
@@ -1531,5 +1490,28 @@ public ResponseEntity<Resource> downloadReport(@RequestParam(required=true) Stri
...
 	}
 	
+	@PostMapping(path = {"/register-promo-pro"})
+	public String registerPromo(@RequestParam String name,
+	                            @RequestParam int subActiveMonths,
+	                            @RequestParam int numberLimit,
+	                            @RequestParam String endTime,
+	                            final RedirectAttributes redirectAttrs) throws ParseException {
+		
+		PromoRepository.Promo promo = new PromoRepository.Promo();
```
vshcherb, line 1500: Better naming PromoCampaign

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/AdminController.java
vshcherb: Just to remind it's a for loop of for loops SQL query, this is definitely not good. It should be 1 query or probably if it is used for information only, it's ok to store inside promocampaign table

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/ApiController.java
```diff
@@ -575,5 +574,13 @@ public String index(HttpServletRequest request, @RequestParam(required = false)
 	public String runLottery(@RequestParam(required = true) String latestHash) throws IOException {
 		return gson.toJson(lotteryPlayService.runLottery(latestHash));
     }
+	
+	@PostMapping(path = {"/promo-add-user"})
+	public ResponseEntity<String> addUser(@RequestParam String promoKey,
+	                                      @RequestParam String userEmail) {
+		synchronized (this) {
```
vshcherb, line 581: this is absolutely the same as public synchronized ResponseEntity<String> addUser(@RequestParam String promoKey,
	                                      @RequestParam String userEmail)

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/PromoService.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+        return ResponseEntity.badRequest().body("Unfortunately we ran out of available promocodes");
+    }
+    
+    private String getLastUsers(PromoCampaignRepository.Promo promoCampaign) {
+        StringBuilder res = new StringBuilder();
+        List<DeviceSubscriptionsRepository.SupporterDeviceSubscription> subscriptions = subscriptionsRepository.findFirst5BySkuOrderByStarttimeDesc("promo_" + promoCampaign.name);
+        for (DeviceSubscriptionsRepository.SupporterDeviceSubscription subscription : subscriptions) {
+            PremiumUsersRepository.PremiumUser user = usersRepository.findByOrderid(subscription.orderId);
```
vshcherb, line 63: add if (subscription.starttime > now() - 1 month) so we don't pull data for years back

PR #638, opened 2023-06-20
fix email text

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/12666

PR #657, opened 2023-07-31
Fix 1449 Route without name

Dima-1:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/osm/RelationTagsPropagation.java
```diff
@@ -152,9 +152,10 @@ private List<RelationRulePropagation> processRelationTags(MapRenderingTypesEncod
 		}
 		return res;
 	}
-	
-	public void handleRelationPropogatedTags(Relation relation, MapRenderingTypesEncoder renderingTypes, OsmDbAccessorContext ctx, 
-			EntityConvertApplyType at) throws SQLException {
+
+	public void handleRelationPropogatedTags(Relation relation, MapRenderingTypesEncoder renderingTypes,
+	                                         OsmDbAccessorContext ctx, EntityConvertApplyType at) throws SQLException {
+		relation.fillEmptyRouteName();
```
vshcherb, line 158: it should be processed after, I think it should be in processRelation method for Map or POI.
Anyway it shouldn't change the tags and use same approach as others
```
Map<String, String> relationTags = ...
```

PR #679, opened 2023-09-12
add approximate for route

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2088

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/RoutingService.java
```diff
@@ -65,6 +62,18 @@ public List<WebGpxParser.Point> updateRouteBetweenPoints(LatLon startLatLon, Lat
         return pointsRes;
     }
     
+    public synchronized List<WebGpxParser.Point> approximateRoute(List<WebGpxParser.Point> points, String routeMode) throws IOException, InterruptedException {
+        List<Location> locations = new ArrayList<>();
+        List<RouteSegmentResult> approximateResult = osmAndMapsService.approximateRoute(points, routeMode);
```
vshcherb, line 67: why this synchronized? no shared resources is used

PR #688, opened 2023-10-11
Add get-styles

alisa911:


alisa911:
Screenshot 2023-10-11 at 16 08 04

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WebService.java
```diff
@@ -0,0 +1,93 @@
...
+import net.osmand.render.RenderingRulesStorage;
+import org.springframework.stereotype.Service;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Service
+public class WebService {
```
vshcherb, line 15: Name is too generic

PR #689, opened 2023-10-11
Add osmand wiki api

Dima-1:


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WikiService.java
```diff
@@ -0,0 +1,93 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.*;
+
+@Service
+public class WikiService {
+	protected static final Log log = LogFactory.getLog(WikiService.class);
+	public static final String WIKIMEDIA_COMMON_SPECIAL_FILE_PATH = "https://commons.wikimedia.org/wiki/Special:FilePath/";
+
+	@Value("${osmand.wiki.location}")
```
ivanPyrohivskyi, line 26: Set to wiki.location full path to *.sqlite file

PR #709, opened 2023-11-09
Fix wikivoyage navigation

Dima-1:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/travel/WikivoyageGenOSM.java
```diff
@@ -237,7 +245,17 @@ public int compare(String s1, String s2) {
...
 	}
-	
+
+	private static void spreadArticleToEverySubregions(OsmandRegions regions, GPXFile gpxFile) {
+		for (WorldRegion region : regions.getWorldRegion().getSubregions()) {
+			WptPt wptPt = new WptPt();
+			LatLon regionCenter = region.getSubregions().get(0).getRegionCenter();
+			wptPt.lat = regionCenter.getLatitude();
+			wptPt.lon = regionCenter.getLongitude();
+			gpxFile.addPoint(wptPt);
```
vshcherb, line 255: why do we add to gpxfile random region center points?

ivanPyrohivskyi: This will add articles without location to all continents:
COUNTRIES=(africa asia europe russia australia-oceania-all central-america north-america south-america)
I this case article will be available in Travel Guides.

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/travel/WikivoyageGenOSM.java
```diff
@@ -237,7 +245,17 @@ public int compare(String s1, String s2) {
 		System.out.println(String.format("Total saved articles: %d", totalArticles));
 		System.out.println(String.format("Empty article: %d no points in article + %d no location page articles (total %d) ", emptyContent, emptyLocation, total));
 	}
-	
+
+	private static void spreadArticleToEverySubregions(OsmandRegions regions, GPXFile gpxFile) {
+		for (WorldRegion region : regions.getWorldRegion().getSubregions()) {
+			WptPt wptPt = new WptPt();
+			LatLon regionCenter = region.getSubregions().get(0).getRegionCenter();
```
ivanPyrohivskyi, line 252: I don't like get(0).something in any code. Add check on nullptr

PR #713, opened 2023-11-17
Fix wikivoyage navigation

Dima-1:


vshcherb (changes requested):
public static final int COMMON_ARTICLE_LAT_LON = -85; // Antarctica LatLon(-85,-85) is  a bad idea

PR #722, opened 2023-11-29
Hotfix for IPv6 vs DownloadServerRegion.matchesIp

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2304

PR #724, opened 2023-12-04
Tools: Random Route Tester beta-version (ready)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18512

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/router/tester/RandomRouteGenerator.java
```diff
@@ -0,0 +1,281 @@
...
+					}
+				}
+			}
+			if (entry.start != null && entry.finish != null) {
+				testList.add(entry);
+			}
+		}
+	}
+
+	private enum randomActions {
```
ivanPyrohivskyi, line 81: Name of enum to uppercase

PR #725, opened 2023-12-05
[HH-SERVER] New routeMode params: hhoff, hhonly (hidden), add server-side hhOnlyLimit, add Elevation data to routing, ...

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18512

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/RoutingService.java
```diff
@@ -102,12 +102,20 @@ private List<WebGpxParser.Point> getStraightLine(double lat1, double lng1, doubl
...
         line.add(lastP);
-        
+
         return line;
     }
-    
-    public void fillRoutingModeParams(RoutingController.RoutingParameter nativeRouting, RoutingController.RoutingParameter nativeTrack, RoutingController.RoutingParameter calcMode,
-                                       RoutingController.RoutingParameter shortWay, Map.Entry<String, GeneralRouter> e, RoutingController.RoutingMode rm) {
+
+	public void fillRoutingModeParams(
```
vshcherb, line 109: formatting

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -160,12 +154,19 @@ public RoutingParameter(String key,  String name, String description, String gro
 	
 	@RequestMapping(path = "/routing-modes", produces = {MediaType.APPLICATION_JSON_VALUE})
 	public ResponseEntity<?> routingParams() {
+		final Set<String> HH_ONLY_FORCED_PROFILES = Set.of("car", "bicycle");
 		Map<String, RoutingMode> routers = new LinkedHashMap<>();
-		RoutingParameter nativeRouting = new RoutingParameter("nativerouting", "Development", 
-				"[Dev] C++ routing", true);
+		RoutingParameter hhRouting = new RoutingParameter("hhrouting", "Development",
```
vshcherb, line 159: 1 parameter only "Do not use HH (routing) instead of 3

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
```diff
@@ -741,6 +746,10 @@ private RoutingContext prepareRouterContext(String routeMode, QuadRect points, R
 				useNativeLib = Boolean.parseBoolean(value);
 			} else if (key.equals("nativeapproximation")) {
 				router.setUseNativeApproximation(Boolean.parseBoolean(value));
+			} else if (key.equals("hhrouting")) {
```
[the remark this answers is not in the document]
RZR-UA: In telegram we've decided to refactor to equalize web routeMode with API routeMode, but later.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -160,12 +154,19 @@ public RoutingParameter(String key,  String name, String description, String gro
 	
 	@RequestMapping(path = "/routing-modes", produces = {MediaType.APPLICATION_JSON_VALUE})
 	public ResponseEntity<?> routingParams() {
+		final Set<String> HH_ONLY_FORCED_PROFILES = Set.of("car", "bicycle");
```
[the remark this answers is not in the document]
RZR-UA: discussed in tg

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -271,62 +296,72 @@ public ResponseEntity<String> routing(@RequestParam String[] points,
...
 	@PostMapping(path = {"/update-route-between-points"}, produces = "application/json")
 	@ResponseBody
-	public ResponseEntity<String> updateRouteBetweenPoints(@RequestParam String start,
-	                                                       @RequestParam String end,
-	                                                       @RequestParam String routeMode,
-	                                                       @RequestParam boolean hasRouting,
-	                                                       @RequestParam int maxDist) throws IOException, InterruptedException {
+	public ResponseEntity<String> updateRouteBetweenPoints(@RequestParam String start, @RequestParam String end,
+	                                                       @RequestParam String routeMode, @RequestParam boolean hasRouting,
+	                                                       @RequestParam(defaultValue = "100") int hhOnlyLimit)
```
vshcherb, line 354: hh only limit should be server side constant

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
```diff
@@ -725,6 +725,11 @@ private RoutingContext prepareRouterContext(String routeMode, QuadRect points, R
 		Map<String, String> paramsR = new LinkedHashMap<String, String>();
 		boolean useNativeLib = DEFAULT_USE_ROUTING_NATIVE_LIB;
 		router.setUseNativeApproximation(false);
+
+		// reset before applying
+		router.USE_HH_ROUTING = false;
```
vshcherb, line 730: USE_HH_ROUTING  should be default true

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
```diff
@@ -793,6 +803,9 @@ public synchronized List<RouteSegmentResult> routing(String routeMode, Map<Strin
 			usedMapList = getReaders(list);
 			
 			RoutingContext ctx = prepareRouterContext(routeMode, points, router, rsc, avoidRoadsIds, usedMapList);
+			if (hhOnlyForce) {
```
vshcherb, line 806: Better to calculate the force switch here - router.USE_ONLY_HH_ROUTING, without passing extra params

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -161,11 +177,15 @@ public RoutingParameter(String key,  String name, String description, String gro
 	@RequestMapping(path = "/routing-modes", produces = {MediaType.APPLICATION_JSON_VALUE})
 	public ResponseEntity<?> routingParams() {
 		Map<String, RoutingMode> routers = new LinkedHashMap<>();
-		RoutingParameter nativeRouting = new RoutingParameter("nativerouting", "Development", 
-				"[Dev] C++ routing", true);
+		RoutingParameter hhRouting = new RoutingParameter("hhrouting", "Development",
+				"[Dev] Use HH (auto)", true);
```
vshcherb, line 181: [Dev] Off HH routing (false)

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -161,11 +177,15 @@ public RoutingParameter(String key,  String name, String description, String gro
...
+		RoutingParameter hhOnly = new RoutingParameter("hhonly", "Development",
```
vshcherb, line 182: Not needed for UI (only for API)

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -161,11 +177,15 @@ public RoutingParameter(String key,  String name, String description, String gro
...
+				"[Dev] Use HH (only)", false);
+		RoutingParameter nativeRouting = new RoutingParameter("nativerouting", "Development",
+				"[Dev] Use C++ (routing)", true);
```
vshcherb, line 185: default false

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: defaultValue here is only to protect against crash if no hhOnlyLimit specified. actually this defaultValue should be never applied because hhOnlyLimit is an always-specified parameter by web-client

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
RZR-UA: USE_HH_ROUTING must be reset here before applying real API parameter which *IS* actually true by default.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
RZR-UA: In this method, router is not defined before method was called, so we need hhOnlyForce param to set the USE_ONLY_HH_ROUTING flag.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: >Not needed for UI (only for API)

Which one isn't needed for the UI?

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: If we set it to default false now, we might overload our server by profile=pedestrian requests !

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: Ok

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
vshcherb: Comment to last line ```RoutingParameter hhOnly = new RoutingParameter("hhonly", "Development",```

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
vshcherb: ? profile=pedestrian - don't understand. [Dev] Use C++ (routing) - default false

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
vshcherb: hhOnlyLimit - shouldn't be passed as parameter

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: 9f8dc4f0c565a20b9232ab11e89018c910865e69

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
RZR-UA: 9f8dc4f0c565a20b9232ab11e89018c910865e69

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: 9f8dc4f0c565a20b9232ab11e89018c910865e69

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: 9f8dc4f0c565a20b9232ab11e89018c910865e69

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: 2094aa805ed8a0378a8eb42bdce82983861f36a0

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
RZR-UA: RoutePlannerFrontEnd router = new RoutePlannerFrontEnd();

`router` is a dynamically created class in this method, so using hhOnlyForce as distinct parameter to change variable in just created class instance is Okay.

It's basically not so good idea to have USE_ONLY_HH_ROUTING (etc!) defined as static, because OsmAndServer is a **potentially multi-threaded** App and should not have any "global" static variables as settings which might be defined independendly...

PR #729, opened 2023-12-13
Try to fix live-updates duplicates

ivanPyrohivskyi:
To issues 
https://github.com/osmandapp/OsmAnd/issues/18153 
https://github.com/osmandapp/OsmAnd/issues/18722

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/diff/ObfDiffMerger.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/diff/ObfDiffMerger.java
@@ -146,13 +146,10 @@ public class ObfDiffMerger {
 				if (commonObj == null) {
 					commonMapData.put(id, mi.adoptMapObject(relObj));
 					cnt++;
-				} else if (deleteId == -1 || !commonObj.containsType(deleteId)) {
+				} else {
 					commonMapData.remove(id);
 					commonMapData.put(id, mi.adoptMapObject(relObj));
 					cnt++;
-				}
-				if (commonObj != null && commonObj.containsType(deleteId)) {
-					System.out.println("Map id:" + relObj.getId() + " has osmand_change=delete tag (" + relObj.toString() + ")");
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/diff/ObfDiffMerger.java
```diff
@@ -145,8 +163,22 @@ private void mergeRelationDiff(String[] args) throws IOException {
...
+					deletedShiftedIds.add(commonObj.getId() >> SHIFT_ID);
+				}
+			}
+		}
+
 		cnt = 0;
 		for (Long id : relRouteData.keys()) {
+			if (deletedShiftedIds.contains(id >> SHIFT_ID)) {
+				continue;
+			}
```
vshcherb, line 181: Below there is a check 
```
	for (Long id : relMapData.keys()) {
				BinaryMapDataObject relObj = relMapData.get(id);
				BinaryMapDataObject commonObj = commonMapData.get(id);
} else if (deleteId == -1 || !commonObj.containsType(deleteId)) {
```
If these ids are not correct and checks are not correct they need to be fixed instead of extra loops

vshcherb:
If these ids are not correct and checks are not correct they need to be fixed instead of extra loops

vshcherb (changes requested):
If these ids are not correct and checks are not correct they need to be fixed instead of extra loops

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/diff/ObfDiffMerger.java
```diff
@@ -118,24 +119,38 @@ private void mergeRelationDiff(String[] args) throws IOException {
 		BinaryMapIndexReader.MapIndex mi = commonObf.getMapIndex();
 		Integer rl = mi.getRule(OSMAND_CHANGE_TAG, OSMAND_CHANGE_VALUE);
 		int deleteId = rl == null ? -1 : rl;
+
+		HashSet<Long> deletedShiftedIds = new HashSet<>();
+		if (deleteId > 0) {
+			for (MapZooms.MapZoomPair mz : commonObf.getZooms()) {
```
vshcherb, line 125: 1. deleted ids could be better inside loop  ```for (MapZooms.MapZoomPair mz : commonObf.getZooms()) {```
2. The main question remains do we understand the reason why we have different ids

PR #732, opened 2023-12-19
[HH-SERVER] Get back applyApproximation online routing option

RZR-UA:


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
```diff
@@ -879,14 +900,15 @@ private List<RouteSegmentResult> onlineRoutingOSRM(String baseurl, RoutingContex
...
+			LOGGER.error(String.format("OSRM empty result (%s)", url));
+			return null;
 		}
 
-		return null;
+		return polyline;
 	}
 
-	private List<RouteSegmentResult> onlineRoutingRescuetrack(String baseurl, RoutingContext ctx,
+	private List<LatLon> onlineRoutingRescuetrack(String baseurl, RoutingContext ctx,
```
vshcherb, line 911: timestamps will be lost better WptPt

RZR-UA:
obsolete

PR #735, opened 2023-12-25
Fix wrong wikivoyage redir (/travel) Issue #18771

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18771

PR #738, opened 2024-01-03
[SMTP_SERVER] Implement SmtpSendGridSender subclass for EmailSenderService

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2335

PR #744, opened 2024-01-19
[HH-SERVER] fix hhoff vs hhonly bug

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/pull/18934

PR #757, opened 2024-02-08
[->rzr-email] Accept `lang` as an API option to use language-specific email templates

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2380

PR #763, opened 2024-02-21
[RandomRouteTester] Calculate distance based on real segment geometry

RZR-UA:


RZR-UA:
Float/double are used in the same way as in calculateTimeSpeed() from RouteResultPreparation

https://github.com/osmandapp/OsmAnd-Issues/issues/2412

PR #768, opened 2024-02-27
[HH-cpp] Add HH-cpp support in OsmAndServer, fix NPE

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2412

PR #772, opened 2024-02-29
fix user api

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2439

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -649,7 +649,33 @@ public ResponseEntity<String> sendCode(@RequestBody EmailSenderInfo data) {
...
 		}
-		return userdataService.sendCode(data.action, data.lang, dev);
+		PremiumUsersRepository.PremiumUser pu = usersRepository.findById(dev.userid);
+		if (pu == null) {
+			return ResponseEntity.badRequest().body("User not found");
+		}
+		return userdataService.sendCode(data.action, data.lang, pu);
+	}
+	
+	@PostMapping(path = {"/auth/send-code-to-new-email"})
```
RZR-UA, line 659: Should be protected against flooding PremiumUsersRepository with tons of emails.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -648,7 +649,33 @@ public ResponseEntity<String> sendCode(@RequestBody EmailSenderInfo data) {
...
+	@ResponseBody
+	public ResponseEntity<String> sendCodeToNewEmail(@RequestBody EmailSenderInfo data, @RequestParam String email) {
```
vshcherb, line 661: Here we don't need ```EmailSenderInfo data``` later again will be same issue that fields to class could be added and affect security

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -663,16 +690,19 @@ public ResponseEntity<String> confirmCode(@RequestBody String code) {
 	
 	@PostMapping(path = {"/auth/change-email"})
 	@ResponseBody
-	public ResponseEntity<String> changeEmail(@RequestBody UserPasswordPost us, HttpServletRequest request) throws ServletException {
-		if (us.username != null) {
-			us.username = us.username.toLowerCase().trim();
-		}
-		if (emailSender.isEmail(us.username)) {
+	public ResponseEntity<String> changeEmail(@RequestBody Map<String, String> credentials, HttpServletRequest request) throws ServletException {
```
vshcherb, line 693: Could you use more transparent way? @RequestBody String username, @RequestBody String token?

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -211,12 +204,17 @@ public ResponseEntity<String> deleteAccount(@RequestParam String token, HttpServ
 
 	@PostMapping(path = { "/auth/activate" }, consumes = "application/json", produces = "application/json")
 	@ResponseBody
-	public ResponseEntity<String> activateMapUser(@RequestBody UserPasswordPost us, HttpServletRequest request)
-			throws ServletException, IOException {
-		ResponseEntity<String> res = userdataService.webUserActivate(us.username, us.token, us.password);
+	public ResponseEntity<String> activateMapUser(@RequestBody Map<String, String> credentials, HttpServletRequest request) throws ServletException {
```
vshcherb, line 207: Map is not better than class.

PR #778, opened 2024-03-07
try fix 16-bit samples are not supported for Horizontal differencing …

alisa911:
…Predictor


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2435

PR #785, opened 2024-03-11
fix api confirm send code

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2439

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -905,43 +906,87 @@ public ResponseEntity<String> confirmCode(String code, PremiumUserDevicesReposit
...
+        // change email
+        usersRepository.delete(tempUser);
+        currentUser.email = username;
+        usersRepository.saveAndFlush(currentUser);
+        request.logout();
+        
+        return ok();
+    }
+    
+    @Scheduled(fixedDelay = 1000 * 60 * 60 * 24)
```
vshcherb, line 980: Not sure this needs to be done so far we can do it manually with review sometimes

vshcherb (approved):
Just remove auto clean up

PR #789, opened 2024-03-13
OsmAndServer calcMode COMPLEX refactor (default router mode fix, hide non-car-derived calcmode from the web)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2472

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
```diff
@@ -880,25 +870,35 @@ private synchronized List<RouteSegmentResult> approximateInternal(RoutingContext
...
-	
+
 
 	private static class RouteParameters {
 		String routeProfile;
 		Map<String, String> routeParams = new LinkedHashMap<String, String>();
 		public RouteParameters(String p) {
 			this.routeProfile = p;
+			final String CAR = "car";
+			Set<String> complexModeProfiles = new HashSet<>();
```
vshcherb, line 881: Doesn't look nice ```RoutingConfiguration.getDefault().getAllRouters()...``` Too many things goes on in constructor

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -212,14 +212,16 @@ public ResponseEntity<?> routingParams() {
...
+						List<RoutingController.RoutingParameter> passParams = "car".equals(e.getKey())
+								? Arrays.asList(hhRouting, nativeRouting, nativeTrack, sepMaps, calcMode)
+								: Arrays.asList(hhRouting, nativeRouting, nativeTrack, sepMaps);
+						routingService.fillRoutingModeParams(passParams, shortWay, e, rm);
 					}
 				} else {
 					rm = new RoutingMode(e.getKey());
 					routers.put(rm.key, rm);
 					routingService.fillRoutingModeParams(
-							Arrays.asList(hhRouting, nativeRouting, nativeTrack, sepMaps, calcMode), shortWay, e, rm);
```
vshcherb, line 222: Here it looks much nicer to configure RoutingParameters, for example calculationmode could always pass to route url but not be visible (as default parameter) in url for web page

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
RZR-UA: fixed (removed because rm became nullable)

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
RZR-UA: As calcMode is nullable now, we should not require calculationmode as mandatory parameter from the web. Just done little refactor after review.

PR #809, opened 2024-03-26
Implement PremiumUserDevice lang/info fields

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2485

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -241,10 +239,19 @@ public ResponseEntity<String> userRegister(@RequestParam(name = "email") String
...
+												 @RequestParam(name = "token", required = true) String token,
+												 @RequestParam(name = "deviceid", required = false) String deviceId,
+												 @RequestParam(name = "brand", required = false) String brand,
+												 @RequestParam(name = "model", required = false) String model,
+												 @RequestParam(name = "lang", required = false) String lang,
+												 @RequestHeader("User-Agent") String userAgent
+	) throws IOException {
 		String accessToken = UUID.randomUUID().toString();
-		return userdataService.registerNewDevice(email, token, deviceId, accessToken);
+		StringBuilder info = new StringBuilder(userAgent == null ? "OsmAnd" : userAgent);
```
vshcherb, line 250: No need user agent (i.e. Version / OsmAnd+) in info

RZR-UA: done

PR #815, opened 2024-04-02
Subway stations on click don't display and suggest subway lanes

Dima-1:


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexTransportCreator.java
```diff
@@ -1143,34 +1143,36 @@ private void mergePlatformsStops(List<Entity> platformsAndStopsToProcess, List<E
...
 			Relation rr = stopAreas.get(EntityId.valueOf(platform));
 			for (Entity stop : stops) {
-				if (stop.getLatLon() == null) {
-					continue;
-				}
 				if (rr != null && stopAreas.get(EntityId.valueOf(stop)) == rr) {
 					replaceStop = stop;
 				}
-				if (MapUtils.getDistance(stop.getLatLon(), loc) < dist) {
+				if (MapUtils.getDistance(stop.getLatLon(), platformLatLon) < dist) {
```
vshcherb, line 1153: 2 ! Checks fo rnull are deleted and here it's expected that it won't crash ??? 
```
if (stop.getLatLon() == null) {
			LatLon loc = platform.getLatLon();

```

PR #816, opened 2024-04-02
Parse 'end' as 'finish' (URL compatibility)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19422

PR #823, opened 2024-04-08
Propose automatedEnvironment skel for Map Creator

RZR-UA:


vshcherb:
This scheme is unlikely going to be supported and understood by other people, so no sense to keep it.
It's better to focus on url support

PR #824, opened 2024-04-08
[4.7] [GpxRouteApproximation] Remove usage of GpxRouteApproximation finalPoints, add setUseNativeApproximation in Map Creator

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2411

PR #840, opened 2024-05-01
[4.7] Implement "Use external timestamps" for RescueTrack

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2342

vshcherb:
4.7 branch!

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
```diff
@@ -923,30 +917,34 @@ private List<RouteSegmentResult> approximateByPolyline(List<LatLon> polyline, St
...
+			route = approximate(ctx, router, props, polyline, null);
 		} finally {
 			unlockReaders(usedMapList);
 		}
 		return route;
 	}
 
-
 	public List<RouteSegmentResult> approximate(RoutingContext ctx, RoutePlannerFrontEnd router,
-			Map<String, Object> props, List<LatLon> polyline) throws IOException, InterruptedException {
```
vshcherb, line 935: Why do we need 2 arrays? Looks very weird @Nullable List<WptPt> sourcePoints is enough

RZR-UA: approximate() method is based on polyline array as source data; there is 4 different sources: Web (gpx OR track points), OSRM (JSON to poliline) and RescueTrack (GPX).

It's not a good idea to spend a lot of time for refactoring conversion from 4 different data sources just to avoid using List<LatLon> in approximate().

PR #846, opened 2024-05-09
Explore – Web

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2516

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WikiService.java
```diff
@@ -58,9 +59,16 @@ public FeatureCollection getImages(String northWest, String southEast) {
...
+		String query = String.format("SELECT id, mediaId, namespace, imageTitle, imgLat, imgLon " +
+				"FROM wikigeoimages WHERE id = %d " +
+				"ORDER BY views DESC LIMIT " + LIMIT_QUERY, id);
+		return getPoiData(null, null, query, "imgLat", "imgLon", null);
+	}
+	
 	public FeatureCollection getWikidataData(String northWest, String southEast, String lang, Set<String> filters) {
 		String filterQuery = filters.isEmpty() ? "" : "AND poitype IN (" + filters.stream().map(s -> "'" + s + "'").collect(Collectors.joining(", ")) + ")";
-		String query = "SELECT id, photoId, photoTitle, catId, catTitle, depId, depTitle, wikiTitle, wikiLang, wikiDesc, wikiArticles, osmid, osmtype, poitype, poisubtype, lat, lon "
+		String query = "SELECT * "
```
vshcherb, line 71: Bad idea, we need to specify what's retrieved. So it's only very specific and transparent

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
```diff
@@ -160,4 +157,20 @@ public ResponseEntity<String> getWikiImages(@RequestParam String northWest, @Req
         FeatureCollection collection = wikiService.getImages(northWest, southEast);
         return ResponseEntity.ok(gson.toJson(collection));
     }
+    
+    @GetMapping(path = {"/get-poi-by-coords"}, produces = "application/json")
+    @ResponseBody
+    public ResponseEntity<String> getPoiByCoords(@RequestParam double lat,
```
vshcherb, line 163: No osm type specified, it's likely possible bug cause osmid could be way/node/relation

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -219,10 +258,38 @@ public SearchUICore.SearchResultCollection searchCitiesByBbox(QuadRect searchBbo
         return res;
     }
     
+    public SearchUICore.SearchResultCollection searchAllPoiByBbox(QuadRect searchBbox, List<BinaryMapIndexReader> mapList) throws IOException {
```
vshcherb, line 261: Is it used?

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -219,10 +258,38 @@ public SearchUICore.SearchResultCollection searchCitiesByBbox(QuadRect searchBbo
...
+        settings.setRegions(osmandRegions);
+        settings.setOfflineIndexes(mapList);
+        
+        Set<String> types = mapPoiTypes.getCategories().stream()
+                .map(PoiCategory::getKeyName)
+                .collect(Collectors.toSet());
+        SearchUICore.SearchResultCollection res = searchWithBbox(searchUICore, settings, searchBbox, types);
+        
+        int attempts = 0;
+        while ((res == null || res.getCurrentSearchResults().isEmpty()) && attempts < 10) {
```
vshcherb, line 280: This is not correct! It shouldn't increase radius osm id should be exact in 10 m radius

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -157,6 +164,38 @@ public PoiSearchResult searchPoi(SearchService.PoiSearchData data) throws IOExce
...
+        SearchResult res = null;
+        try {
+            List<OsmAndMapsService.BinaryMapIndexReaderReference> mapList = getMapsForSearch(bbox, searchBbox);
+            if (!mapList.isEmpty()) {
+                usedMapList = osmAndMapsService.getReaders(mapList, null);
+                SearchUICore.SearchResultCollection resultCollection = searchAllPoiByBbox(searchBbox, usedMapList);
+                if (resultCollection != null) {
+                    List<SearchResult> searchResults = resultCollection.getCurrentSearchResults();
+                    if (!searchResults.isEmpty()) {
+                        res = searchResults.stream()
```
vshcherb, line 183: 1. No need to store all results to retrieve 1 object. Probably better way should be used via accept / filter
2. No need to use search at all as we just need to read POI object

PR #848, opened 2024-05-13
Implement routing/approximation enums, useExternalTimestamps for Web gpx/route, refactor web options

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2342

PR #860, opened 2024-06-10
Try to fix device-register API requests (code 400)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2570

PR #862, opened 2024-06-10
Another try to fix 400 error for device-register

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2570

PR #863, opened 2024-06-11
fix wiki photos

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/17949

PR #873, opened 2024-06-26
Hotfix HH-to-BRP fallback NORMAL to COMPLEX mode

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/pull/20116/files

PR #875, opened 2024-07-03
Fix generation of negative ids (JOSM "draft" maps problem)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2556

vshcherb:
Needs to be reviewed and thorougly read by @ivanPyrohivskyi to be merged. probably it's correct. Think what kind of maps we also generate with negative id to cehck

ivanPyrohivskyi:
Checked with negative IDs:
- test turn lanes, works, but size increase on 2Mb (90/92 Mb)
- synthetic test rendering, visual only
- SRTM generation (Albania)

ivanPyrohivskyi (approved):

PR #876, opened 2024-07-03
Implement generate-hh-routing.sh and documentation for custom HH-generation

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2556

vshcherb:
not sure value of thhis script
1. we're not going to use it
2. we're not going to support it
3. we're not going to package it

we have only utilities.sh, if they need these script and fine with it so they can support it you can send them. this script too complicated anyway better not to have it - less code is better

PR #878, opened 2024-07-07
Hotfix misplaced removal of anonymous iOS devices

RZR-UA:
~15% of OsmAnd iOS users send device-register API requests with empty deviceId (see the snippets below).

Current implementation of delete-same-device code **always removes ALL devices** from Cloud in case if user has more than one iOS Cloud device with disabled sendAnonymousAppUsageData option.

To reproduce the bug, we need 2 iOS devices with the same Cloud account, both should have anonymous option setup. Register new device on 2nd "anonymous" device will remove 1st one.

Hotfix is proposed. Logging included.

```
- (NSString *) getUserIosId {
...
    if (![settings.sendAnonymousAppUsageData get])
        return @""; // might be empty
...
```

```
@implementation OARegisterUserCommand
...
NSString *deviceId = OsmAndApp.instance.getUserIosId;
params[@"deviceid"] = deviceId; // might be empty
```


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2604

PR #879, opened 2024-07-10
Use interpolateEmptyElevationWpts for Web (Navigate and Plan route)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19142

PR #891, opened 2024-08-21
Draft upgrade to SpringBoot 3.0.0 (a lot of to do)

RZR-UA:


vshcherb:
Might be useful in future as upgrade will need anyway in 1-2 years

PR #893, opened 2024-08-23
remove poi filters for web categories

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/460

PR #895, opened 2024-08-24
Use OsmAnd-shared in/around OsmGpxWriteContext

RZR-UA:
Using GpxFile, GpxUtilities, GpxTrackAnalysis from Kotlin lib:

DownloadOsmGPX 100% (tested on corresponding Jenkins job)
OsmGpxWriteContext 100% (tested on convert-gpx-to-obf utility)

MapApiController minimally
OsmAndMapsService minimally

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
@@ -2,6 +2,9 @@ package net.osmand.server.controllers.user;
 
 import java.io.*;
 
+import okio.GzipSource;
+import okio.Okio;
+
 import static net.osmand.server.api.services.FavoriteService.FILE_TYPE_FAVOURITES;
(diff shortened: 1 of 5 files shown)


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18866

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -2,6 +2,9 @@
 
 import java.io.*;
 
+import okio.GzipSource;
+import okio.Okio;
```
ivanPyrohivskyi, line 6: Where yet we use Okio ?

ivanPyrohivskyi (approved):

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
RZR-UA: Kotlin/shared-lib

PR #908, opened 2024-09-13
add filter wikidata by zoom

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/497

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WikiService.java
```diff
@@ -81,16 +85,36 @@ public FeatureCollection getImagesById(long id, double lat, double lon) {
 		return new FeatureCollection(features.toArray(new Feature[0]));
 	}
 	
-	public FeatureCollection getWikidataData(String northWest, String southEast, String lang, Set<String> filters) {
+	public FeatureCollection getWikidataData(String northWest, String southEast, String lang, Set<String> filters, int zoom) {
 		String filterQuery = filters.isEmpty() ? "" : "AND poitype IN (" + filters.stream().map(s -> "'" + s + "'").collect(Collectors.joining(", ")) + ")";
```
RZR-UA, line 89: Please avoid using **user-defined** `filters` inside string-formed SQL query. It is vulnerable for SQL injections. Using any strings as "'" + string "'" is an awful approach of forming SQL queries. Use SQL placeholders instead.

vshcherb:
Need to test wikidata coords filter first

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WikiService.java
```diff
@@ -45,6 +45,10 @@ public class WikiService {
 
 	private static final int LIMIT_QUERY = 1000;
 	private static final int LIMITI_QUERY = 25;
+	
+	private static final Map<Integer, Set<String>> EXCLUDED_POI_TYPES_BY_ZOOM = Map.of(14, Set.of("shop"));
+	private static final Map<Integer, Set<String>> EXCLUDED_POI_SUBTYPES_BY_ZOOM = Map.of(14, Set.of("artwork", "commercial", "memorial", "monument"));
```
vshcherb, line 50: Some artwork / memorial could be very popular https://www.wikidata.org/wiki/Q185382

PR #919, opened 2024-10-01
Support [\r\n] in file names (platform-independend) + Fix Invalid characters (CR/LF) in header Content-Disposition + fix web list-files

RZR-UA:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -392,6 +393,11 @@ public ResponseEntity<String> listFiles(@RequestParam(required = false) String n
 							: userdataService.getInputStream(uf);
 					if (in != null) {
 						GPXFile gpxFile = GPXUtilities.loadGPXFile(new GZIPInputStream(in));
+						if (gpxFile.error != null) {
+							LOG.error("list-files: ignore corrupted-gpx-file (" + uf.id + ") " + uf.name);
```
ivanPyrohivskyi, line 397: Add info about user to log

ivanPyrohivskyi (approved):

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
RZR-UA: userid added

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20934

PR #920, opened 2024-10-02
Fix NPE for UNKNOWN FileType download-backup

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2714

RZR-UA:
outdated

PR #934, opened 2024-10-18
Fix merge tagGroups for country obf

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/2993
Depends from https://github.com/osmandapp/OsmAnd/pull/21079


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexPoiCreator.java
```diff
@@ -1054,4 +1076,71 @@ private PoiType getPropagatedPoiType(String subtype) {
 		propagatedPoiTypeCache.put(subtype, poiType);
 		return poiType;
 	}
+
+	private String insertMergedTaggroups(Amenity amenity) throws SQLException {
```
vshcherb, line 1080: Not thoroughly checked, needs to be tested

PR #940, opened 2024-10-24
Fix issue with timeout by reconnect

vshcherb:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2745

PR #942, opened 2024-10-26
Fix problem with HTML_COMMENTS replaceAll and empty "headers"

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2690

RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2690

PR #944, opened 2024-10-29
Web - display routes and filter by types / years

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2708

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/OsmGpxController.java
```diff
@@ -0,0 +1,202 @@
...
+	                                        @RequestParam String maxlon) {
+		if (!config.osmgpxInitialized()) {
+			return ResponseEntity.ok("OsmGpx datasource is not initialized");
+		}
+		cleanupCache();
+		String conditions = "";
+		conditions += " and m.maxlat >= " + minlat;
+		conditions += " and m.minlat <= " + maxlat;
+		conditions += " and m.maxlon >= " + minlon;
+		conditions += " and m.minlon <= " + maxlon;
```
RZR-UA, line 69: Please validate user-defined `minlat, maxlat, minlon, maxlon` by converting them to Float and then back to String. You should never allow invalidated user-defined variables in SQL-queries. Additionally, validate or use placeholders for `date` and `activities`.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/OsmGpxController.java
```diff
@@ -0,0 +1,202 @@
...
+				for (String tagOr : tagsOr) {
+					if (t) {
+						conditions += " or ";
+					}
+					conditions += " lower(m.tags::text) like '%" + tagOr.trim().toLowerCase() + "%'";
+					t = true;
+				}
+				conditions += ")";
+			}
+		}
```
RZR-UA, line 87: Activity-selection should be rewritted with new approach (activity field).

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/OsmGpxController.java
```diff
@@ -0,0 +1,202 @@
...
+
+		// filter by year
+		if (!Algorithms.isEmpty(year)) {
+			conditions += " and extract(year from m.date) = " + year;
+		}
+
+		String query = "SELECT m.id, f.data AS bytes, m.name, m.description, m.\"user\", m.date, m.tags " +
```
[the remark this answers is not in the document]
RZR-UA: (seems it is done)

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/server/osmgpx/DownloadOsmGPX.java
```diff
@@ -216,12 +214,197 @@ public static void main(String[] args) throws Exception {
...
+			}
+		}
+		return null;
+	}
+
+	private static Map<String, List<String>> createActivitiesMap(String rootPath) {
+		File activityFile = new File(rootPath, "resources/poi/activities.json");
+		if (!activityFile.exists()) {
+			return Collections.emptyMap();
+		}
```
RZR-UA, line 339: Please avoid calling this method for each track. Just do it one-off at the start of the utility.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/server/osmgpx/DownloadOsmGPX.java
```diff
@@ -216,12 +214,197 @@ public static void main(String[] args) throws Exception {
...
+				return "foot";
+			} else if (avgSpeed <= 25) {
+				return "cycling";
+			} else if (avgSpeed <= 150) {
+				return "driving";
+			} else {
+				return "aviation";
+			}
+		}
+		return null;
```
RZR-UA, line 405: It should never return null or the main SELECT cycle will loop forever.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/server/osmgpx/DownloadOsmGPX.java
```diff
@@ -216,12 +214,197 @@ public static void main(String[] args) throws Exception {
...
+				return "other";
+			} else if (avgSpeed <= 12) {
+				return "foot";
+			} else if (avgSpeed <= 25) {
+				return "cycling";
+			} else if (avgSpeed <= 150) {
+				return "driving";
+			} else {
+				return "aviation";
+			}
```
RZR-UA, line 403: Please add validation, for instance: analysis.hasSpeedInTrack() && ... to prevent mark track as "aviation" on case of NaN-like values.

RZR-UA (changes requested):

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/server/osmgpx/DownloadOsmGPX.java
```diff
@@ -389,20 +389,22 @@ private String analyzeActivityFromGpx(byte[] bytes) {
...
+					return "driving";
+				} else if (avgSpeed > 150) {
+					return "aviation";
+				} else {
+					return "other";
+				}
 			}
 		}
-		return null;
+		return "other";
```
RZR-UA, line 412: It's better to return "error" is gpxFile is null.

RZR-UA (approved):

PR #948, opened 2024-10-31
Implement no-conditional-routing option for Web Routing

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2748

PR #949, opened 2024-10-31
fix multithreading issues

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2677

vshcherb (changes requested):
Make 	public synchronized boolean validateAndInitConfig() throws IOException {

PR #955, opened 2024-11-06
Autodetect routeCalculationTime by LatLon (start point), allow set fixed time in URL (routeCalculationTime:ms)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2748

PR #958, opened 2024-11-08
use new parsing for old api

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/545

PR #963, opened 2024-11-11
Fix bugs around loadGPXFile and fix API download-obf (bug with multiple-as-one gpx files and "0" shield text)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2277

PR #968, opened 2024-11-13
Clarify MSG_LONG_DIST message

RZR-UA:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -55,7 +55,7 @@
 @Controller
 @RequestMapping("/routing")
 public class RoutingController {
-	public static final String MSG_LONG_DIST = "Sorry, in our beta mode max routing distance is limited to ";
+	public static final String MSG_LONG_DIST = "Unable to process your route. Please limit the distance to ";
```
vshcherb, line 58: Issue we still request to limit to 100 km. Message should be dependent on the error whether we ask to limit to 1000km (if distance is over > 5000 km) or 100km (when we don't have index or limit. manually)

PR #1000, opened 2024-12-24
speed up vector tiles

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2845

RZR-UA (approved):

Issue #1002, opened 2024-12-26
Live OBF download URL changed and breaks older OsmAndMapCreator releases


cmoffroad:
Looks like the 'Download Live' URL for live OBF updates has changed, as older OsmAndMapCreator versions can’t access these links anymore:  
https://download.osmand.net/download?aosmc=true&self=true&file=Thailand_asia_24_12_00.obf.gz  

I can’t upgrade to JDK 17 yet, which is required for the latest OsmAndMapCreator build:  
https://osmand.net/docs/versions/map-creator/  

Is there an older version that works with JDK 16 and supports the updated URLs?

vshcherb:
problem with url we need to double check what's new url

cmoffroad:
> Another url works https://download.osmand.net/download?aosmc=true&file=Thailand_asia_24_12_00.obf.gz

Redirects to and returns 404 Not Found: https://dl8.osmand.net/download?aosmc=true&file=Thailand_asia_24_12_00.obf.gz&region=thailand_asia&self=true

RZR-UA:
The original URL is fixed and should work now:

`https://download.osmand.net/download?aosmc=true&self=true&file=Thailand_asia_24_12_00.obf.gz`

PR #1005, opened 2024-12-28
Parse MapCreator URL (aosmc=true instead of =yes)

RZR-UA:
1) Fix wrong server selection for URL such as `https://download.osmand.net/download?aosmc=true&file=Thailand_asia_24_12_00.obf.gz`

`aosmc=true` is incorrect and it should be set to "yes" to select right servers in load balancer code

2) Additionally, the URL is changed to correct (**possibly, the 2nd commit is enough to fix issue**)


RZR-UA:
https://github.com/osmandapp/OsmAnd-tools/issues/1002

PR #1006, opened 2024-12-28
hotfix url

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-tools/issues/1002

PR #1014, opened 2025-01-06
Fix relation tags propagation

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/21420


ivanPyrohivskyi:
......

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/osm/RelationTagsPropagation.java
```diff
@@ -167,7 +167,7 @@ public void handleRelationPropogatedTags(Relation relation, MapRenderingTypesEnc
 			for (RelationMember ids : relation.getMembers()) {
 				PropagateEntityTags entityTags = getPropogateTagForEntity(ids.getEntityId());
 				for (RelationRulePropagation p : lst) {
-					String sortKey = RELATION_SORT_TAG + p.relationGroupKeyString;
+					String sortKey = RELATION_SORT_TAG + p.relationGroupKeyString + ":" + p.relationGroupValueString;
```
vshcherb, line 170: This is doesn't look correct 
```
sortKey = RELATION_SORT_TAG + p.relationGroupKeyString
```
It's done on purpose, so international are displayed on top https://github.com/osmandapp/OsmAnd/issues/9236#issuecomment-643817113.

I think this commit breaks sorting between iwn, nwn,... So it needs to fully be tested

PR #1021, opened 2025-01-20
Auth of OsmAnd Cloud devices using Web credentials

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/issues/4252

PR #1022, opened 2025-01-22
Fix MapCreator build (move GpxFile to Kotlin lib)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2846

RZR-UA:
approved by @alex-osm

PR #1025, opened 2025-01-24
Climbing polygons

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/21248


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexVectorMapCreator.java
```diff
@@ -1339,4 +1347,41 @@ private boolean isClockwiseBroken(Map<String, String> tags, Way e, List<Node> si
         return clockwiseAfter != clockwiseBefore;
     }
 
+    private static MultipolygonBuilder createClimbingMultipolygon(Entity e, OsmDbAccessorContext ctx) throws SQLException {
```
vshcherb, line 1350: 1. Do we need statics here? 
2. Optional to think: It's might to big class in the end if we can move this logic to MultipolygonBuilder or special place?

ivanPyrohivskyi: Moved to MultipolygonBuilder, depended PR https://github.com/osmandapp/OsmAnd/pull/21796

PR #1040, opened 2025-02-11
Allow multiple from/to members in no_entry/no_exit relations

klonfish:
This fixes osmandapp/OsmAnd#21855 by looping over the list of members and adding multiple `RestrictionInfo` to the corresponding ways if that is valid for the given restriction type. For relations that only allow one way for each role, it keeps the current behavior of using only the first one.

I tested it in some scenarios and it seems to work correctly.

While I was at it, i also added two small commits to improve the code of the function a little bit.

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
@@ -623,10 +623,10 @@ public class IndexRouteCreator extends AbstractIndexPartCreator {
 					type = MapRenderingTypes.RESTRICTION_ONLY_STRAIGHT_ON;
 				}
 				if (type != -1) {
-					ctx.loadEntityRelation((Relation) e);
-					Collection<RelationMember> fromL = ((Relation) e).getMembers("from"); //$NON-NLS-1$
-					Collection<RelationMember> toL = ((Relation) e).getMembers("to"); //$NON-NLS-1$
-					Collection<RelationMember> viaL = ((Relation) e).getMembers("via"); //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
```diff
@@ -623,27 +628,39 @@ private void indexHighwayRestrictions(Entity e, OsmDbAccessorContext ctx) throws
...
+					if (!toL.isEmpty()) {
+						for (RelationMember from : fromL) {
+							if (from.getEntityId().getType() == EntityType.WAY) {
+								if (!highwayRestrictions.containsKey(from.getEntityId().getId())) {
+									highwayRestrictions.put(from.getEntityId().getId(), new ArrayList<>());
+								}
+
+								List<RestrictionInfo> rdList = highwayRestrictions.get(from.getEntityId().getId());
+								for (RelationMember to : toL) {
+									RestrictionInfo rd = new RestrictionInfo();
```
vshcherb, line 644: This code needs to be simplififed, used to be 5 inners ifs, but now 4 ifs + 2 fors too complex

klonfish: Thanks for your feedback. I totally agree.
I guess this would be a good use case for guard clauses to return immediately when one of the member lists is empty. This would eliminate two nesting levels. Additionally, the checks for `val` (currently line 592) and `type` (currently line 630) could also be written as guard clauses, further reducing the nesting of the whole function.
Would this approach be okay for you? If so, I would add a commit containing only these refactorings.

ivanPyrohivskyi (approved):

PR #1041, opened 2025-02-12
fix route types from web gpx file

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/661

RZR-UA (approved):

PR #1045, opened 2025-02-21
add read access for public share files

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2527

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/ShareFileService.java
```diff
@@ -330,4 +330,31 @@ public ResponseEntity<String> saveSharedFile(String name, String type, String ne
...
+		for (ShareFileRepository.ShareFilesAccess access : accessList) {
+			if (access.getUser().id == userId) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	@Transactional
+	public void createAccess(ShareFileRepository.ShareFile shareFile, PremiumUserDevicesRepository.PremiumUserDevice dev, String nickname) {
```
RZR-UA, line 345: 1) (Security) What happens when the owner changes the file access to non-public? Will the file still be accessible because of this "created" access?

2) Consider better naming: createReadAccess? createPublicReadAccess?

3) Why nickname is always null? It will be used in the future?

RZR-UA (approved):

PR #1048, opened 2025-02-27
Generate `route_shortlink_tiles` tag

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/16808

PR #1054, opened 2025-03-05
OsmGpxWriteContext fixes (distance in km, search by filename if the track name is changed)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2277

PR #1055, opened 2025-03-06
Overwrite OSM routes distance (distanceKmFormat)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/16808

PR #1062, opened 2025-03-19
Merge and cropp tiles during contour generation

ihoryakovenko:
https://github.com/osmandapp/OsmAnd/issues/21738

### DIFF
--- a/obf-generation/contours/V4/make-contour-tile-mt.sh
+++ b/obf-generation/contours/V4/make-contour-tile-mt.sh
@@ -203,13 +203,29 @@ process_tiff ()
                     neighbors+=("$neighbor_filename")
 
                     if [ $dlat -eq 0 ] && [ $dlon -eq -1 ]; then
-                        left_exists=1
+			    if [[ "$orig_lon_prefix" == "E" ]]; then
+                        	left_exists=1
+			    else
+				right_exists=1
+			    fi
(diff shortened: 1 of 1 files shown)


obf-generation/contours/V4/make-contour-tile-mt.sh
```diff
@@ -138,7 +138,162 @@ process_tiff ()
...
+                    else
+                        lon_prefix="E"
+                    fi
+                    new_lon="${new_lon#-}"  # Remove the negative sign
+                fi
+        
+                formatted_lat=$(printf "%0${lat_digits}d" "$new_lat")
+                formatted_lon=$(printf "%0${lon_digits}d" "$new_lon")
+
+                neighbor_filename="${indir}${lat_prefix}${formatted_lat}${lon_prefix}${formatted_lon}.tif"
```
xmd5a2, line 198: Slash (**/**) should be added after ${indir} because earlier it is removed
`indir=${1%/*}`

xmd5a2 (commented):
Contours should be generated even if tiff tile doesn't have neighbors at all. It is a common situation in world ocean.
Tile N00E009 is calculated incorrectly (see orange small contours at top and bottom of tile)
Screenshot_2025-03-21_17-18-58

obf-generation/contours/V4/make-contour-tile-mt.sh
ihoryakovenko: Fixed

PR #1071, opened 2025-04-02
fix bug with 2 identical rte

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-iOS/issues/4466

PR #1089, opened 2025-04-14
Fixed subscription validation check

Corwin-Kh:


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserSubscriptionService.java
```diff
@@ -72,7 +72,7 @@ public String checkOrderIdPremium(String orderid) {
 					s = revalidateiOSSubscription(s);
 				}
 			}
-			if (s.valid == null || !s.valid.booleanValue()) {
```
vshcherb, line 75: s.expiretime == null? separate case?

PR #1097, opened 2025-04-18
Fix id generation

ivanPyrohivskyi:
Related commit with bug: https://github.com/osmandapp/OsmAnd-tools/commit/104b0ba170b1c915d4bf41f679b38745b6546866


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/OsmDbCreator.java
```diff
@@ -153,7 +155,7 @@ public long convertId(Entity e) {
 				EntityType entityType = i.getEntityId().getType();
 				if (i.getEntityId().getType() != EntityType.RELATION) {
 					Long newId ;
-					if (!addGeoHash) {
+					if (!addGeoHash && useSimpleIds) {
```
vshcherb, line 158: useSimpleIds everywhere equals !addGeoHash

PR #1103, opened 2025-04-24
speed up get-wiki-data api

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2964

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WikiService.java
```diff
@@ -204,17 +204,31 @@ public FeatureCollection getWikidataData(String northWest, String southEast, Str
...
+		String query = buildWikidataQuery(langPriority, showAll, filterQuery);
+
+		return getPoiData(northWest, southEast, query, filterParams, "lat", "lon", langPriority);
+	}
+
+	private String buildWikidataQuery(List<String> langPriority, boolean showAll, String filterQuery) {
+
+		String langList = langPriority.stream()
+				.map(l -> "'" + l.trim() + "'")
+				.collect(Collectors.joining(", ", "[", "]"));
```
RZR-UA, line 218: Vulnerable to SQL injections. Please rewrite.

vshcherb: If there is no way around you need to check that lang is <= 3 chars and only consists of 'a' - 'z' (lowercase)

RZR-UA (approved):

PR #1114, opened 2025-05-05
add city for poi search result

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/765

RZR-UA (approved):

PR #1126, opened 2025-05-14
rename info file

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/744

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WebUserdataService.java
```diff
@@ -381,6 +387,10 @@ public ResponseEntity<String> renameFile(String oldName, String newName, String
 	}
 
 	private File renameGpxTrack(PremiumUserFilesRepository.UserFile file, String newName) throws IOException {
+		String preparedName = newName.substring(0, newName.lastIndexOf('.'));
+		if (preparedName.contains("/")) {
+			preparedName = preparedName.substring(preparedName.lastIndexOf('/') + 1);
+		}
```
RZR-UA, line 393: Could be simplified.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/WebUserdataService.java
```diff
@@ -366,9 +366,15 @@ public ResponseEntity<String> renameFile(String oldName, String newName, String
 				}
 				ResponseEntity<String> res = userdataService.uploadFile(zipFile, dev, newName, type, System.currentTimeMillis());
 				if (res.getStatusCode().is2xxSuccessful()) {
+					boolean renamed = renameInfoFile(oldName, newName, dev);
+					if (!renamed) {
+						return ResponseEntity.badRequest().body("Error rename info file!");
```
RZR-UA, line 371: What if the `.info` file does not exist?

PR #1146, opened 2025-05-23
Implement --unsubscribe-from instead of --update_postfix_bounced

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2989

PR #1157, opened 2025-06-05
Fix duplicated RDO types after SRTM processing

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21977

PR #1161, opened 2025-06-09
Possible fix for sub names

vshcherb:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/AdminController.java
```diff
@@ -1192,73 +1193,42 @@ public void processRow(ResultSet rs) throws SQLException {
...
+		s.defPriceEurMillis = subBaseData.defaultPriceEurMillis();
 	}
 	
 	
 	public static class SubscriptionMonthReport {
 		public String date;
 	}
 	
-	public static enum SubAppType {
+	public enum SubAppType {
```
vshcherb, line 1213: osmand, osmand_plus is not a purchase provider? ios, huawei, amazon, fastspring

PR #1166, opened 2025-06-11
Overhaul IpLocationService with GeoIP legacy files

RZR-UA:


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/IpLocationService.java
```diff
@@ -1,81 +1,72 @@
...
+	public static final String LON = "lon"; // used by Web
+	public static final String LATITUDE = "latitude"; // used by Android and iOS
+	public static final String LONGITUDE = "longitude"; // used by Android and iOS
+	public static final String COUNTRY_CODE = "country_code"; // used by OsmAnd-telegram (NL)
+	public static final String COUNTRY_NAME = "country_name"; // used by tools (Netherlands)
+	public static final String CITY = "city"; // used by tools (Amstelveen)
 
-    private static final Log LOGGER = LogFactory.getLog(IpLocationService.class);
+	public static final String GeoIPCityV4 = "/usr/share/GeoIP/GeoIPCity.dat";
+	public static final String GeoIPCityV6 = "/usr/share/GeoIP/GeoIPCityv6.dat";
```
ivanPyrohivskyi: @Value("${osmand.heightmap.location}")
	String geotiffTiles;

RZR-UA: bd753ebd5

PR #1182, opened 2025-06-20
add sendOsmAndSpecialGiftEmail

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -72,7 +72,26 @@ public boolean sendPromocodesEmails(String mailTo, String templateId, String pro
...
+						"OsmAnd Team";
+
+		boolean ok = new EmailSenderTemplate()
+				.template("Subject: " + subject + "\n" + body)
+				.to(email)
+				.send()
+				.isSuccess();
+
+		LOGGER.info("sendOsmAndSpecialGiftEmail to: " + shorten(email) + " (" + ok + ")");
+	}
```
RZR-UA: Please avoid hardcoding the subject and body. Create a corresponding template in `web-server-config/templates/email/cloud/`. If possible, follow languages. Refer to sendOsmAndCloudRegistrationEmail and sendOsmAndCloudWebEmail.

PR #1197, opened 2025-07-10
Load nativeLib from NATIVE_LIBRARY_PATH (optional)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22938

PR #1227, opened 2025-08-15
remove hidden false from gpx

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/848

RZR-UA (approved):

PR #1232, opened 2025-08-19
use categoryimages_final

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/23204

RZR-UA:
categoryimages_final would not be used

PR #1238, opened 2025-08-26
Max token fix

Gateway07:
1. safe_max_tokens is made more smart
2. Number of completions is increased to 2 attempts
3. Adding LLM_TIMEOUT=360 and MODEL_TOP_P=1


RZR-UA:
Please elaborate on all changed parameters and their impact on the result.

PR #1242, opened 2025-08-28
Discard old user link to allow register new cloud account

vshcherb:


RZR-UA (approved):

alisa911:
https://github.com/osmandapp/web/issues/972

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -186,23 +186,22 @@ public ResponseEntity<String> userUpdateOrderid(@RequestParam(name = "email", re
...
+                                               HttpServletRequest request) {
 		// allow to register only with small case
 		email = email.toLowerCase().trim();
 		CloudUser pu = usersRepository.findByEmailIgnoreCase(email);
 		if (!email.contains("@")) {
 			logErrorWithThrow(request, ERROR_CODE_EMAIL_IS_INVALID, "email is not valid to be registered");
 		}
 		if (pu != null) {
-			if (!login) {
+			if (Boolean.FALSE.equals(login)) {
```
RZR-UA: Previous logic: `login == null` is equivalent to `login == false`. Why does the logic change to `== false` only?

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -286,13 +285,12 @@ public ResponseEntity<String> userRegister(
...
 
-	private void discardPreviousAccountOrderId(String orderid, String currentEmail, HttpServletRequest request) {
+	protected void discardPreviousAccountOrderId(int newUserId, String orderid, String currentEmail, HttpServletRequest request) {
 		CloudUser previousUser =
 				usersRepository.findFirstByOrderidAndEmailNotIgnoreCaseOrderByIdAsc(orderid, currentEmail);
 		if (previousUser != null) {
 			if (DISCARD_ANOTHER_USER_PAYMENT_IF_NEW_REGISTERED) {
-				previousUser.orderid = null;
-				usersRepository.saveAndFlush(previousUser);
+				userSubService.relinkPurchasesToNewAccount(previousUser, newUserId);
```
RZR-UA, line 294: Please log the orderId here, along with the oldUserId and newUserId. We should be able to find records in logs using the orderId.

alisa911 (approved):

PR #1243, opened 2025-09-01
Search-test v3.0 (testing-nocode)

Gateway07:
1. Multi-modules for custom JS functions.
2. Testing mode for visual testing of test-case data.
3. No-code mode to configure test-case without JS function.

https://github.com/osmandapp/OsmAnd-Issues/issues/2905#issuecomment-3240850273

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/SearchTestRepositoryConfiguration.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/SearchTestRepositoryConfiguration.java
@@ -16,15 +16,12 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.FilterType;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
 import org.springframework.jdbc.core.JdbcTemplate;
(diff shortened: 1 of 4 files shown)


java-tools/OsmAndServer/src/main/java/net/osmand/server/SearchTestRepositoryConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+            HibernateProperties hibernateProperties) {
+        // Derive vendor properties from Spring Boot configuration (includes ddl-auto, naming, etc.)
+        Map<String, Object> vendorProps = new java.util.HashMap<>(
+                hibernateProperties.determineHibernateProperties(
+                        jpaProperties.getProperties(), new HibernateSettings()
+                )
+        );
+        // Force SQLite dialect for this EMF
+        vendorProps.put("hibernate.dialect", "net.osmand.server.StrictSQLiteDialect");
+        vendorProps.put("hibernate.hbm2ddl.auto", "update");
```
RZR-UA: Wrong way round.

PR #1250, opened 2025-09-03
Do not delete-same-device for "not-the-same" model (server hotfix for duplicated iOS aid/deviceId)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/3087

PR #1278, opened 2025-09-25
Set restriction via to the end of idTable

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/23403

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
@@ -1489,7 +1489,7 @@ public class IndexRouteCreator extends AbstractIndexPartCreator {
                 ids.add(id);
             }
         }
-        if (!restrictionVia.isEmpty() && !ids.isEmpty()) {
+        if (!ids.isEmpty() && restrictionVia.contains(ids.get(0))) {
             // set restrictionVia to the end of idTable (val.viaWay != 0)
             for (int i = 1; i < ids.size(); i++) {
                 if (!restrictionVia.contains(ids.get(i))) {


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
```diff
@@ -589,6 +589,9 @@ private void deleteRouteTreeFiles(String rTreeMapIndexNonPackFileName, String rT
 	private void indexHighwayRestrictions(Entity e, OsmDbAccessorContext ctx) throws SQLException {
 		if (e instanceof Relation && "restriction".equals(e.getTag(OSMTagKey.TYPE))) { //$NON-NLS-1$
 			String val = e.getTag("restriction"); //$NON-NLS-1$
+            if (e.getId() == 19565585L) {
```
vshcherb: ?

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
```diff
@@ -1472,9 +1475,31 @@ public static void writeBinaryMapBlock(rtree.Node parent, Rect parentBounds, RTr
 		wc.wayMapIds.clear();
 		wc.wayMapIdsCache.clear();
 		wc.pointMapIds.clear();
-		for (int i = 0; i < parent.getTotalElements(); i++) {
-			if (e[i].getElementType() == rtree.Node.LEAF_NODE) {
-				long id = e[i].getPtr();
+        List<Long> restrictionVia = new ArrayList<>();
```
vshcherb: LLM Review: 
2. Performance and Data Structures
Multiple Loops: The code iterates through all elements to build two lists (ids and restrictionVia), and then iterates through restrictionVia again to modify ids. This can be simplified.

LinkedList vs. ArrayList: You are adding to the end of the ids list (ids.add(id)), for which ArrayList is generally faster. The choice of LinkedList was likely to make ids.remove(0) an efficient O(1) operation. However, since the logic that relies on it is flawed, the justification for using LinkedList disappears. ArrayList is a better default choice unless you have a specific, frequent need to remove from the beginning or middle of a list.

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexRouteCreator.java
```diff
@@ -1472,9 +1475,31 @@ public static void writeBinaryMapBlock(rtree.Node parent, Rect parentBounds, RTr
...
+                        if(rd.viaWay != 0) {
+                            restrictionVia.add(rd.viaWay);
+                        }
+                    }
+                }
+                ids.add(id);
+            }
+        }
+        // set restrictionVia to the end of idTable (val.viaWay != 0)
+        for (long id : restrictionVia) {
```
vshcherb: very weird loop seems very ineffecient

vshcherb: LLM suggested:
```
List<Long> restrictionVia = new ArrayList<>();
List<Long> ids = new LinkedList<>();

// 1. Populate the lists as you did before.
....
// 2. This is the new, simple logic to fix the first element.
Set<Long> viaWaySet = new HashSet<>(restrictionVia);
if (!ids.isEmpty() && viaWaySet.contains(ids.get(0))) {
    for (int i = 1; i < ids.size(); i++) {
        if (!viaWaySet.contains(ids.get(i))) {
            Collections.swap(ids, 0, i);
            break;
        }
    }
}
```

PR #1289, opened 2025-10-02
search-test-back-v2.4

Gateway07:
- **Adding rerun, res_id link fix**
- **byTag fix**
- **Comparison report fix**
- **Dup count**
- **Dup minDistance fix**

https://github.com/osmandapp/OsmAnd-Issues/issues/2905#issuecomment-3346747430

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/searchtest/MapDataObjectFinder.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/searchtest/MapDataObjectFinder.java
@@ -92,6 +92,7 @@ public class MapDataObjectFinder {
 		if (result.location == null || result.localeName == null || result.file == null)
 			return new Result(ResultType.Error, null, place, result);
 
+		final String[] byTag = {null};
 		final BinaryMapDataObject[] byDist = {null}, byId = {null};
 		QuadRect quad = MapUtils.calculateLatLonBbox(result.location.getLatitude(), result.location.getLongitude(), 100);
 		BinaryMapIndexReader.SearchRequest<BinaryMapDataObject> request = BinaryMapIndexReader.buildSearchRequest(
@@ -119,7 +120,7 @@ public class MapDataObjectFinder {
 						String value = tags.get(tag);
 						boolean matches = value != null && value.startsWith(result.localeName);
 						if (matches)
-							row.put("by_tag", tag + "=" + value);
+							byTag[0] = tag + "=" + value;
 						return matches;
 					}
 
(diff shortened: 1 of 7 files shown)


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/searchtest/MapDataObjectFinder.java
```diff
@@ -112,7 +119,10 @@ public boolean publish(BinaryMapDataObject obj) {
 							return false;
 
 						String value = tags.get(tag);
-						return value != null && value.startsWith(result.localeName);
+						boolean matches = value != null && value.startsWith(result.localeName);
+						if (matches)
```
vshcherb, line 123: By tat is incorrect it should check distance and all tags not 1

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/searchtest/MapDataObjectFinder.java
```diff
@@ -83,10 +89,11 @@ public Result[] find(List<SearchResult> searchResults, long datasetId) {
 		return new Result[] {firstResult, null};
 	}
 
-	public Result getMapDataObject(SearchResult result, long expectedOsmId, int place) {
+	public Result getMapDataObject(SearchResult result, long expectedOsmId, int place, Map<String, Object> row) {
```
vshcherb, line 92: 1. We always read file again ? this is very slow, we don't even check if we have already id or not

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/searchtest/MapDataObjectFinder.java
```diff
@@ -55,13 +56,13 @@ private Map<String, String> getTags(BinaryMapDataObject obj) {
...
 
-	public Result[] find(List<SearchResult> searchResults, long datasetId) {
+	public Result[] find(List<SearchResult> searchResults, long datasetId, Map<String, Object> row) {
 		int resPlace = 1;
 		Result firstResult = null, firstByTag = null, firstByDist = null;
 		for (SearchResult res : searchResults) {
 			Object wt = res.objectType;
 			if (wt != null && !"LOCATION".equals(wt.toString())) {
-				Result currResult = getMapDataObject(res, datasetId, resPlace);
+				Result currResult = getMapDataObject(res, datasetId, resPlace, row);
```
vshcherb, line 65: This is way too many reads without checking even bbox

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/searchtest/MapDataObjectFinder.java
vshcherb: 2. Row is only used by_tag which not needed

PR #1293, opened 2025-10-03
Add ID for building in address section

ivanPyrohivskyi:


vshcherb:
Unnecessary data for users

PR #1309, opened 2025-10-16
create new intersection name for search result

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-iOS/issues/4909

PR #1317, opened 2025-10-20
fix types param for generateFiles

alisa911:


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -438,7 +438,8 @@ public ResponseEntity<String> listFiles(@RequestParam(name = "deviceid", require
 		if (dev == null) {
 			return userdataService.tokenNotValidError();
 		}
-		UserFilesResults res = userdataService.generateFiles(dev.userid, name, allVersions, false, type);
+		Set<String> types = type != null ? Set.of(type) : Collections.emptySet();
```
RZR-UA: type.split(...) ?

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -252,6 +252,15 @@ public UserdataController.UserFilesResults generateFiles(int userId, String name
...
 	}
 
+	public Set<String> parseFileTypes(String types) {
+		return (types != null)
+				? Arrays.stream(types.split(","))
+				.map(String::trim)
+				.filter(s -> !s.isEmpty())
+				.collect(Collectors.toCollection(HashSet::new))
+				: Collections.emptySet();
+	}
```
RZR-UA, line 267: overkill

PR #1392, opened 2025-12-29
Implement simple location sharing

vshcherb:


java-tools/OsmAndServer/src/main/java/net/osmand/server/WebSecurityConfiguration.java
```diff
@@ -95,17 +95,24 @@ public class WebSecurityConfiguration {
...
 		private static final long serialVersionUID = -881322456618342435L;
-		CloudUserDevice userDevice;
+		private final CloudUserDevice userDevice;
+		private final CloudUser pu;
 
-		public OsmAndProUser(String username, String password, CloudUserDevice pud,
+		public OsmAndProUser(String username, String password, CloudUser pu, CloudUserDevice pud,
 				List<GrantedAuthority> authorities) {
 			super(username, password, authorities);
+			this.pu = pu;
```
vshcherb, line 100: Force relogin

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -105,14 +138,41 @@ public static class EmailSenderInfo {
 		public String lang;
 	}
 
-	private CloudUserDevice checkToken(int deviceId, String accessToken) {
+	private static class CachedInfoDevice {
+		final CloudUserDevice device;
+		final long timestamp;
+		public CloudUser user;
```
vshcherb: Check situations - if token is deleted but still cached

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/SubscriptionInterceptor.java
```diff
@@ -2,21 +2,82 @@
...
+	}
+
     @Override
     public Message<?> preSend(@NotNull Message<?> message, @NotNull MessageChannel channel) {
         StompHeaderAccessor accessor = StompHeaderAccessor.wrap(message);
+        
+        // Propagate SecurityContext from HTTP session to WebSocket message
+        // This ensures Principal (Authentication) is available for access control
+        SecurityContext securityContext = SecurityContextHolder.getContext();
+        Authentication authentication = securityContext.getAuthentication();
```
RZR-UA: May authentication be null?

java-tools/OsmAndServer/src/main/java/net/osmand/server/WebSecurityConfiguration.java
```diff
@@ -314,7 +317,7 @@ CorsConfigurationSource corsConfigurationSource() {
 				"https://test.osmand.net", "https://osmbtc.org", "http://localhost:3000"));
 		configuration.setAllowCredentials(true);
 		configuration.setAllowedMethods(Arrays.asList(CorsConfiguration.ALL));
-		configuration.setAllowedHeaders(Arrays.asList("Content-Type"));
+		configuration.setAllowedHeaders(Arrays.asList("Content-Type", "Authorization", "alias", "password", "deviceId"));
```
RZR-UA: Do not use non-natural (from the point of view of the HTTP protocol) names for headers. Use X-Alias, X-Password, X-Device-Id, etc. In addition, I don't think it's a good idea to list variable names in the headers. At least, add a comment about why and where the non-standard headers might be used.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/SubscriptionInterceptor.java
```diff
@@ -30,18 +91,184 @@ public Message<?> preSend(@NotNull Message<?> message, @NotNull MessageChannel c
...
+        // Check if user is an active sharer
+        // Get deviceId from header if specified
+        long deviceId = 0;
+        String deviceIdHeader = headers.getFirstNativeHeader("deviceId");
+        if (deviceIdHeader != null && !deviceIdHeader.isEmpty()) {
+            try {
+                deviceId = Long.parseLong(deviceIdHeader);
+            } catch (NumberFormatException e) {
+                // Invalid deviceId format - ignore
+            }
```
RZR-UA: Use Algorithms.parseLongSilently()

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsController.java
```diff
@@ -17,25 +17,43 @@ public class UserTranslationsController {
 	@Autowired
 	private UserTranslationsService userTranslationsService;
 
-	// One time call (subscription) returns map with TRANSLATION_ID
+
 	@MessageMapping("/translation/create")
-	public Object createTranslation(SimpMessageHeaderAccessor headers, Principal principal) {
+	public Object createTranslation(@Payload(required = false) String password,
+	                                SimpMessageHeaderAccessor headers,
+	                                Principal principal) {
```
RZR-UA: What is the source of the String password for this API?

If it is

```        stompClient.publish({
            destination: '/app/translation/create',
            body: password || ''
        });
```

So why is only the password used as a request body?
What will you do if you need another variable(s) in the future?

Consider using JSON format to exchange parameters in the body/reply.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsController.java
```diff
@@ -47,46 +65,99 @@ public void whoami(SimpMessageHeaderAccessor headers, Principal principal) {
...
+
+	private boolean isValidTranslationId(String translationId) {
+		if (translationId == null || translationId.isEmpty()) {
+			return false;
+		}
+		if (translationId.length() > 128) {
+			return false;
+		}
+		// Allow only alphanumeric characters and safe characters
+		return translationId.matches("^[a-zA-Z0-9_-]+$");
```
RZR-UA: Please define this regexp as public final. It is used twice in the code.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsController.java
```diff
@@ -17,25 +17,43 @@ public class UserTranslationsController {
...
 		CloudUser user = userTranslationsService.getUser(principal, headers);
 		if (user == null) {
 			return null;
 		}
-		return userTranslationsService.createTranslation(user, null, headers);
+		UserTranslation translation = userTranslationsService.createTranslation(user, null, headers);
+
+		if (isValidPassword(password)) {
+			userTranslationsService.setTranslationPassword(translation, password);
+		}
```
RZR-UA: Please add `else if (password != null && !password.isEmpty()) { error }` because somebody might use a password such as `"   {}   "` or just `"       "` and we just ignore it.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsController.java
```diff
@@ -47,46 +65,99 @@ public void whoami(SimpMessageHeaderAccessor headers, Principal principal) {
...
+		return translationId.matches("^[a-zA-Z0-9_-]+$");
+	}
 
+	private boolean isValidPassword(String password) {
+		if (password == null) {
+			return false;
+		}
+		String trimmed = password.trim();
+		return !trimmed.isEmpty() && !trimmed.equals("{}");
+	}
```
RZR-UA: `isValidPassword` seems somewhat confusing. isValid for Password implies that the password is valid and its owner is validated. Consider better naming, such as !isEmptyPassword(). Additionally, password trimming is not the best approach. Spaces are legal in passwords, including at the beginning and end.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -73,10 +84,24 @@ public class UserTranslationsService {
...
     
     private final Environment environment;
+    
+    private final BCryptPasswordEncoder passwordEncoder = new BCryptPasswordEncoder();
 
     public UserTranslationsService(Environment environment) {
         this.environment = environment;
     }
+
+    public void setTranslationPassword(UserTranslation translation, String plainPassword) {
```
RZR-UA: Please, never ever use **plain**Password in Java code.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -160,7 +203,71 @@ public UserTranslation createTranslation(CloudUser user, String translationId, S
...
+		String deviceIdHeader = headers.getFirstNativeHeader("deviceId");
+		if (deviceIdHeader == null || deviceIdHeader.isEmpty()) {
+			return 0;
+		}
+		try {
+			return Long.parseLong(deviceIdHeader);
+		} catch (NumberFormatException e) {
+			LOG.warn("Invalid deviceId format: " + deviceIdHeader);
+			return 0;
+		}
```
RZR-UA: use Algorithms please

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -160,7 +203,71 @@ public UserTranslation createTranslation(CloudUser user, String translationId, S
...
+	}
+
+	private long calculateExpireTime() {
+		return System.currentTimeMillis() + DEFAULT_SHARING_DURATION_MS;
+	}
+
+	private long countUserTranslations(int userId) {
+		long count = 0;
```
RZR-UA: int

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -178,26 +295,136 @@ public void load(UserTranslation ust, SimpMessageHeaderAccessor headers) {
 	}
 
 	// Sharing management
+
+	public void recordVerifiedAccess(String translationId, long userId) {
+		if (translationId != null) {
+			UserTranslation translation = activeTranslations.get(translationId);
+			if (translation != null) {
+				translation.getVerifiedUsers().add((int)userId);
```
RZR-UA: Please use one type for userId. int or long, but one type everywhere. Avoid **(int)**userId

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -178,26 +295,136 @@ public void load(UserTranslation ust, SimpMessageHeaderAccessor headers) {
...
+			// Already sharing, just update expire time
+			updateSharerExpireTime(existingSharer);
+			return;
+		}
+
+		if (deviceId > 0) {
+			long totalDevicesForUser = countUserDevicesInTranslation(ust, user.id);
+			if (totalDevicesForUser >= MAX_DEVICES_PER_USER_PER_TRANSLATION) {
+				String error = "Maximum number of devices (" + MAX_DEVICES_PER_USER_PER_TRANSLATION + 
+					") per user per translation reached";
```
RZR-UA: Use String.format in a way used previously in this commit.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -505,4 +598,10 @@ public ResponseEntity<String> confirmCode(@RequestBody MapApiController.UserPass
 		}
 		return ResponseEntity.badRequest().body("Please enter valid email");
 	}
+
+	@Scheduled(fixedRate = 1, timeUnit = TimeUnit.HOURS) // Run every hour
+	public void clearExpiredTokens() {
+		long now = System.currentTimeMillis();
+		cacheByDeviceId.values().removeIf(c -> (now - c.timestamp) > CACHE_TTL);
```
RZR-UA: CACHE_TTL = 15 minutes. How do we clear the 15-minute TTL in the 60-minute scheduler?

Please could you make a separate request for all cacheByDeviceId features to be reviewed with @ivanPyrohivskyi?

RZR-UA (changes requested):
Please follow the comments.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/SubscriptionInterceptor.java
alisa911: It can return null when no user is authenticated. i check it

alisa911:
## Security Techniques Added

**Password-Protected Room Access via JWT Tokens:**
- JWT Bearer tokens for password-authenticated room access
- Token-to-room binding validation (token must match specific translationId)
- Time-limited tokens (30 minutes validity)
- Token type verification (ROOM token type)

**Rate Limiting:**
- Failed attempt tracking per WebSocket session
- Maximum 5 failed attempts per session
- 1-minute rate limit window
- Automatic reset on successful access

**Device-Based Access Control:**
- Device ID validation via `X-Device-Id` header
- Per-device sharing restrictions (max 5 devices per user per translation)
- Device-specific access matching for active sharers
- Flexible matching: deviceId = 0 allows any device, specific deviceId restricts to that device only

**Verified Users Cache:**
- In-memory cache of verified users per translation
- Avoids repeated password/token checks after initial verification
- Supports anonymous users (id = -1) in verified list

**Access Control Flow:**
- JWT token validation on SUBSCRIBE command
- Fallback to session-based permissions if no token
- Token validation only on subscription, not on every message
- Access granted via token → user added to verifiedUsers cache

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/WebSocketConfig.java
```diff
@@ -45,8 +51,20 @@ public TaskScheduler heartbeatScheduler() {
     }
 	
 	@Override
-	public void registerStompEndpoints(StompEndpointRegistry registry) {
-		registry.addEndpoint("/osmand-websocket");
+	public void registerStompEndpoints(@NotNull StompEndpointRegistry registry) {
+		boolean isProduction = environment.acceptsProfiles(Profiles.of("production"));
```
RZR-UA: Have you tested this `isProduction` in our environment? What happens for http://localhost:... and http://test.osmand.net/ws-test.html ?

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/WebSocketConfig.java
```diff
@@ -45,8 +51,20 @@ public TaskScheduler heartbeatScheduler() {
...
+		
+		if (isProduction) {
+			// In production, only allow connections from osmand.net domains over HTTPS/WSS
+			registry.addEndpoint("/osmand-websocket")
+				.setAllowedOriginPatterns("https://*.osmand.net")
```
RZR-UA: https://*.osmand.net does match https://www.osmand.net but might not for https://osmand.net

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/SubscriptionInterceptor.java
```diff
@@ -235,34 +284,12 @@ private boolean isUserAllowed(String translationId, Principal principal, SimpMes
...
-                    // Remember verified user for this translation only
-                    translation.getVerifiedUsers().add(cloudUser.id);
-                }
-                return matches;
-            } catch (Exception e) {
-                LOG.warn("Error verifying password for translation " + translationId + ": " + e.getMessage());
-                return false;
-            }
+            LOG.debug("Password-protected translation " + translationId + " requires JWT Bearer token");
+            return false;
```
RZR-UA: Always return false for Password-protected translation? Weird.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/RateLimitService.java
```diff
@@ -0,0 +1,80 @@
+package net.osmand.server.ws;
```
RZR-UA: RateLimitService.java - useless, better remove

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -340,42 +340,52 @@ public String generateRoomToken(String translationId, int userId, String alias)
...
 		
-		String passwordHash = translation.getPassword();
-		// If room has no password, allow access
-		if (passwordHash == null || passwordHash.isEmpty()) {
-			return generateRoomToken(translationId, userId, alias);
-		}
+		// Dummy BCrypt hash for timing attack prevention when translation doesn't exist
+		// This ensures password verification always takes similar time
+		// Using a valid BCrypt hash format that will never match any password
+		final String DUMMY_HASH = "$2a$10$N9qo8uLOickgx2ZMRZoMyeIjZAgcfl7p92ldGxad68LJZdL17lhWy";
```
RZR-UA: Useless. Remove and simplify code, please.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -0,0 +1,658 @@
...
+        // Sends to: /user/{sessionId}/queue/updates
+        SimpMessageHeaderAccessor header = SimpMessageHeaderAccessor.create();
+        header.setSessionId(sessionId);
+        header.setLeaveMutable(true);
+        template.convertAndSendToUser(sessionId, QUEUE_USER_UPDATES, payload, header.getMessageHeaders());
+    }
+
+    
+    public CloudUser getUser(Principal principal, SimpMessageHeaderAccessor headers) {
+		boolean production = environment.acceptsProfiles(Profiles.of("production"));
```
RZR-UA: `environment.acceptsProfiles(Profiles.of("production"))` please test it with the local server, builder, and main. We never used this approach to detect the production server.

java-tools/OsmAndServer/src/main/java/net/osmand/server/ws/UserTranslationsService.java
```diff
@@ -0,0 +1,658 @@
...
+		}
+		String nickname = getNickname(user);
+		ust.getPendingShareRequests().put(user.id, nickname);
+		saveTranslationToRedis(ust);
+		Map<String, Object> data = new HashMap<>();
+		data.put("translationId", ust.getId());
+		data.put("userId", user.id);
+		data.put("nickname", nickname);
+		notifyUser((int) ust.getOwner(), USER_UPD_TYPE_SHARE_REQUEST, data);
+	}
```
RZR-UA, line 465: I'm not sure that we need to "ask for the permission", at all. The translation ID is unique enough to serve as "permission automatically granted". Consider removing useless features such as request/grant/deny share.

PR #1476, opened 2026-03-27
poi_improvement

Gateway07:
- **new perf metrics are added**
- **subApiMetricsEnabled flag is added**
- **balanced objects per block writing**
- **index tokens for bloom filter fixing**
- **add metrics for reporting**
- **add token suffix only to bloom**
- **fix summary report**
- **fix MAX_SUBBLOCK_SIZE**
- **fix SUMMARY_REPORT**
- **add OVERSIZED-BLOOM_REPORT**
- **fix OVERSIZED-BLOOM_REPORT**
- **fix reportedBloomTokens**
- **fix reportedBloomTokens**
- **add SPECIAL_REPORT**
- **fix SPECIAL_REPORT**
- **remove MAX_SUBBLOCK_SIZE**
- **remove special report**
- **fix bloomTokens**


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/BinaryMapIndexWriter.java
```diff
@@ -1734,6 +1768,117 @@ public void writePoiTagGroups(PoiCreatorTagGroups tagGroups) throws IOException
...
+		TreeSet<String> sortedSuffixes = new TreeSet<>();
+		Map<T, Set<String>> suffixesByObject = new LinkedHashMap<>();
+		for (T object : objects) {
+			Set<String> objectSuffixes = new LinkedHashSet<>();
+			suffixesByObject.put(object, objectSuffixes);
+			for (String token : tokenSupplier.apply(object)) {
+				if (!token.startsWith(prefix) || token.length() <= prefix.length()) {
+					continue;
+				}
+				String suffix = Normalizer.normalize(token.substring(prefix.length()), Normalizer.Form.NFC);
```
ivanPyrohivskyi, line 1845: It is not good that prefix can be broken. Perhaps better to fix SearchCoreFactory splitSearchNames. In this case can be avoid call Normalizer.normalize(...) and prefix will be with right length.
Or I misunderstood you code?

Gateway07: That final normalization is still useful because String.substringing after a valid boundary does not guarantee NFC.

vshcherb (approved):

PR #1509, opened 2026-04-30
Add new controller get-color-palette

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/543

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -444,6 +444,22 @@ public void getFile(HttpServletResponse response, HttpServletRequest request,
...
+		CloudUserFilesRepository.UserFile userFile = userdataService.getUserFile(
+				"color-palette/user_palette_default.txt", "FILE", null, dev);
+		if (userFile == null) {
+			return ResponseEntity.ok("");
+		}
+		try (InputStream bin = new GZIPInputStream(userdataService.getInputStream(dev, userFile))) {
+			return ResponseEntity.ok(new String(bin.readAllBytes(), java.nio.charset.StandardCharsets.UTF_8));
+		}
+	}
+
```
RZR-UA, line 462: Unnecessary API. Better to use user_palette_default.txt directly from resources, with a possible conversion from TXT to JSON for use with JS (as we do in resources/generated/...).

alisa911: This is not a common resource file, but a custom user file

RZR-UA (approved):

PR #1510, opened 2026-04-30
Search favorites or tracks by name.

Dima-1:
https://github.com/osmandapp/web/issues/1065


alisa911 (commented):
This is the wrong approach for the web. It complicates the search API and adds new dependencies. Why search for tracks on the server if the user's track list already exists on the web?

PR #1525, opened 2026-05-19
Add generatePoiCategoriesJson

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/543

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/server/utilities/StyleDataExtractor.java
```diff
@@ -265,6 +266,84 @@ private static List<RenderingRule> getRules(RenderingRule rule, List<RenderingRu
...
+            sb.append("      \"icons\" : [\n");
+            sb.append("        ").append(iconsLine).append("\n");
+            sb.append("      ]\n");
+            sb.append("    }");
+            if (i < entries.size() - 1) sb.append(",");
+            sb.append("\n");
+        }
+
+        sb.append("  }\n");
+        sb.append("}");
```
RZR-UA: Please avoid using StringBuilder with "pretty-fake" formatting. Try to compile the JSON object and just dump it as a String.

RZR-UA (changes requested):

PR #1530, opened 2026-05-26
Fix favorites group update status shown as add instead of modify

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/24700

PR #1552, opened 2026-06-19
Stream run summary and drop result_json from runs list

alisa911:


alisa911:

PR #1561, opened 2026-06-28
Add analyze-run operation summing file sizes from a previous run

alisa911:


alisa911:

PR #1571, opened 2026-07-01
Fix unsupported renderingParams for v1 tiles (LLM)

RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2845

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/VectorTileController.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/VectorTileController.java
@@ -174,7 +174,7 @@ public class VectorTileController {
 		Map<String, String> props = new TreeMap<>();
 		for (RenderingRuleProperty property : vectorStyle.storage.PROPS.getPoperties()) {
 			String value = renderingParams.get(property.getAttrName());
-			if (!Algorithms.isEmpty(value)) {
+			if (!Algorithms.isEmpty(value) && isSafeRenderingPropValue(value)) {
 				props.put(property.getAttrName(), value);
 			}
 		}
@@ -189,6 +189,10 @@ public class VectorTileController {
 			renderingProps.append(key).append("=").append(value);
 		});
 		return renderingProps.toString();
+	}
+
+	private boolean isSafeRenderingPropValue(String value) {
+		return !value.contains(",") && !value.contains("=") && !value.contains(";");
 	}
 
 	private String getRenderingCacheKey(String renderingProps) {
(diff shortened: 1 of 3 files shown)


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/VectorTileController.java
```diff
@@ -160,13 +167,47 @@ public ResponseEntity<String> getTileInfo(@PathVariable String style, @PathVaria
...
+			renderingProps.append(key).append("=").append(value);
+		});
+		return renderingProps.toString();
+	}
+
+	private String getRenderingCacheKey(String renderingProps) {
+		if (Algorithms.isEmpty(renderingProps)) {
+			return "";
+		}
+		return "rp-" + Integer.toUnsignedString(renderingProps.hashCode(), 36);
```
alisa911, line 202: hashCode used as cache discriminator (collision → wrong tile served)

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/VectorTileController.java
```diff
@@ -160,13 +167,47 @@ public ResponseEntity<String> getTileInfo(@PathVariable String style, @PathVaria
...
 	}
 
-	public VectorMetatile getMetaTile(VectorStyle vectorStyle, int z, int x, int y, String interactiveKey) {
+	private String getRenderingProps(VectorStyle vectorStyle, Map<String, String> renderingParams) {
+		if (renderingParams == null || renderingParams.isEmpty()) {
+			return "";
+		}
+		Map<String, String> props = new TreeMap<>();
+		for (RenderingRuleProperty property : vectorStyle.storage.PROPS.getPoperties()) {
+			String value = renderingParams.get(property.getAttrName());
```
alisa911, line 176: rendering-param values are not sanitized before injection into the props string

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/VectorTileController.java
RZR-UA: hashCode is good enough for props key

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/VectorTileController.java
RZR-UA: 5f596e50c

alisa911 (approved):

PR #1580, opened 2026-07-06
Add wikidata to address section

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/25331


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexAddressCreator.java
```diff
@@ -1081,7 +1078,9 @@ private Map<String, String> getOtherNames(Entity e) {
 				prefix = "";
 			} else if(t.startsWith("loc_name")){
 				prefix = "";
-			}
+			} else if (t.equals("wikidata")) {
```
vshcherb: Not correct see langAttributes and constant NAME_PLACE_ATTR how used

PR #1582, opened 2026-07-07
Fix former-pro users never matching in notify-inactive-users

alisa911:


alisa911:

PR #1597, opened 2026-07-16
Tag filter

Dima-1:
https://github.com/osmandapp/OsmAnd/issues/24809


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -1365,6 +1359,40 @@ public Map<String, List<String>> searchPoiCategories(String locale) {
 		return res;
 	}
 
+	public Map<String, Object> checkTagsVisibility(Map<String, String> tags) {
```
alisa911: Rename the method, a check method should return a bool. Make tag clearer about what tags these are. infoFilter is not actually a filter, and visible is an adjective. tags-visibility translates as "visibility flags".

The unified filtering logic should be in a method in osmand java.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -1365,6 +1359,40 @@ public Map<String, List<String>> searchPoiCategories(String locale) {
...
+		if (tags == null || tags.isEmpty()) {
+			return Collections.emptyMap();
+		}
+		AdditionalInfoBundle infoFilter = new AdditionalInfoBundle(MapPoiTypes.getDefault(), tags);
+		Map<String, Object> visible = new HashMap<>();
+		infoFilter.getFilteredLocalizedInfo().forEach((key, value) -> {
+			if (!infoFilter.shouldDisplayKey(key) && !key.equals(NAME)) {
```
alisa911: This special case for NAME is dead, because a name has no poiType and gets dropped later at the poiType null check, so add a test that name reaches the result.

Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/93eb808c1f30b0a99ca2aa6fe82424626ca7f837

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -1365,6 +1359,40 @@ public Map<String, List<String>> searchPoiCategories(String locale) {
...
+			if (value instanceof String str) {
+				if (infoFilter.isKeyToSkip(key)) {
+					return;
+				}
+				vl = str;
+			}
+			PoiType poiType = infoFilter.getPoiAdditionalType(key, vl);
+			if (poiType == null) {
+				poiType = infoFilter.getPoiAdditionalType(key.replaceAll(":", "_"), vl);
+			}
```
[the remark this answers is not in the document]
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/93eb808c1f30b0a99ca2aa6fe82424626ca7f837

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -1414,6 +1408,91 @@ public Map<String, List<String>> searchPoiCategories(String locale) {
 		return res;
 	}
 
+	public List<Map<String, Object>> getVisibleTags(Map<String, String> tags) {
+		if (tags == null || tags.isEmpty()) {
+			return Collections.emptyList();
+		}
+		AdditionalInfoBundle infoFilter = new AdditionalInfoBundle(MapPoiTypes.getDefault(), tags);
+		Map<String, Object> visible = new HashMap<>();
```
alisa911: Use LinkedHashMap here to keep a stable tag order

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -1414,6 +1408,91 @@ public Map<String, List<String>> searchPoiCategories(String locale) {
...
+		infoFilter.getFilteredLocalizedInfo().forEach((key, value) -> {
```
alisa911: getFilteredLocalizedInfo() is called twice (here and below). Store it in a local variable and iterate once

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -1414,6 +1408,91 @@ public Map<String, List<String>> searchPoiCategories(String locale) {
...
+			if (!infoFilter.shouldDisplayKey(key)) {
+				return;
+			}
+			String vl = null;
+			if (value instanceof String str) {
+				if (infoFilter.isKeyToSkip(key) || key.equals("note")) {
```
alisa911: Of the old client-side shouldSkipKey denylist (name, subway_region, note, lang_yes, *route*), only "note" is reproduced — please confirm the rest are excluded via shouldDisplayKey/poiType, otherwise they'll now leak into the panel.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
```diff
@@ -1414,6 +1408,91 @@ public Map<String, List<String>> searchPoiCategories(String locale) {
...
+				Map<String, Object> entry = new HashMap<>();
+				entry.put("key", key);
+				entry.put("value", mainEntry.value());
+				entry.put("lang", mainEntry.lang());
+				List<Map<String, Object>> otherLangs = entries.stream()
+						.filter(e -> e != mainEntry)
+						.map(e -> {
+							Map<String, Object> otherLengsMap = new HashMap<>();
+							otherLengsMap.put("key", e.key());
+							otherLengsMap.put("value", e.value());
```
alisa911: otherLangsMap

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
```diff
@@ -141,6 +141,13 @@ public ResponseEntity<String> getPoiCategories(@RequestParam String locale) {
             return ResponseEntity.badRequest().body("Error get poi categories!");
         }
     }
+
+    @PostMapping(path = {"/get-visible-tags"}, produces = "application/json")
+    @ResponseBody
+    public ResponseEntity<String> getVisibleTags(@RequestBody Map<String, String> tags) {
```
alisa911: get-visible-tags describes the operation inaccurately, consider filter-visible-tags

alisa911 (commented):
Replace the hand-built maps with string keys in groupLocalizedTags with a small record (VisibleTag) serialized by gson to drop the "key"/"value"/"lang"/"otherLangs"/"localizations" literals and the typo risk, and move the "note" literal to a named constant.

record VisibleTag(String key, String value, String lang, List<VisibleTag> otherLangs) {}

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/964006d4045ff059d629aa7eb3fc36e060a1cbf2

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/SearchService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/964006d4045ff059d629aa7eb3fc36e060a1cbf2

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+				continue;
+			}
+			List<VisibleTag> entries = new ArrayList<>();
+			for (AmenityRowData child : row.collapsableRows) {
+				int idx = child.key.indexOf(':');
+				entries.add(idx >= 0
+						? new VisibleTag(child.key.substring(0, idx), child.value, child.key.substring(idx + 1))
+						: new VisibleTag(child.key, child.value, null));
+			}
+			VisibleTag mainEntry = entries.stream().filter(e -> e.lang() != null).findFirst().orElse(null);
```
alisa911: The "main" localized value is just the first entry with a lang — map
iteration order, not the user's language. Android picks the preferred
locale here. Should the endpoint accept a lang parameter and choose
accordingly?

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
 		}
 		return resultCollection;
 	}
+
+	public List<VisibleTag> filterVisibleTags(Map<String, String> tags) {
```
alisa911: filterVisibleTags doesn't really filter — it builds, sorts and groups the
visible tags. getVisibleTags would be a more honest name (it also matches
the bundle method it wraps).

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+		if (tags == null || tags.isEmpty()) {
+			return Collections.emptyList();
+		}
+		AdditionalInfoBundle infoFilter = new AdditionalInfoBundle(MapPoiTypes.getDefault(), tags);
```
alisa911: Two things here: infoFilter holds an AdditionalInfoBundle, not a filter —
infoBundle would be clearer. And MapPoiTypes.getDefault() bypasses the
per-locale instances this same service manages in getMapPoiTypes() —
better to use one source of poi types.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+		List<AmenityRowData> infoRows = infoFilter.getVisibleTags(false);
```
alisa911: Minor: "rows" is UI wording, on the server these are just tag entries —
visibleRows or tagEntries reads better. Also getVisibleTags(false)
silently drops the "note" tag — worth a short comment why, or a parameter.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+		AmenityRowsBuilder.sortInfoRows(infoRows);
```
alisa911: sortInfoRows compares name as a tie-breaker, but name is always null here
(it's only set on Android by fillRow). So ties are ordered differently
than on Android — probably fine, but worth being aware of.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+		return groupLocalizedTags(infoRows);
+	}
+
+	public record VisibleTag(String key, String value, String lang, List<VisibleTag> otherLangs) {
```
alisa911: Nit: nested types are usually declared at the top or bottom of the class,
not between methods.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+		public VisibleTag(String key, String value, String lang) {
+			this(key, value, lang, null);
+		}
+	}
+
+	private String poiTypeGroupValue(AmenityRowData row) {
```
alisa911: Nit: poiTypeGroupValue actually joins poi type keys with a separator —
joinPoiTypeKeys would say what it does.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+		for (PoiType pt : row.collapsablePoiTypes) {
+			if (!sb.isEmpty()) {
+				sb.append(Amenity.SEPARATOR);
+			}
+			sb.append(pt.getKeyName());
+		}
+		return sb.toString();
+	}
+
+	private List<VisibleTag> groupLocalizedTags(List<AmenityRowData> rows) {
```
alisa911: groupLocalizedTags is really a mapper to the response DTO (grouping is
only one branch) — toVisibleTags would be clearer.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+				entries.add(idx >= 0
+						? new VisibleTag(child.key.substring(0, idx), child.value, child.key.substring(idx + 1))
+						: new VisibleTag(child.key, child.value, null));
+			}
+			VisibleTag mainEntry = entries.stream().filter(e -> e.lang() != null).findFirst().orElse(null);
+			if (mainEntry != null) {
+				List<VisibleTag> otherLangs = entries.stream()
+						.filter(e -> e != mainEntry)
+						.collect(Collectors.toList());
+				result.add(new VisibleTag(row.key, mainEntry.value(), mainEntry.lang(),
```
alisa911: Nit: every entry in otherLangs repeats the parent key — {value, lang}
would be enough and would shrink the response.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,65 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+						otherLangs.isEmpty() ? null : otherLangs));
+			} else if (!entries.isEmpty()) {
```
alisa911: This "else if (!entries.isEmpty())" can't be false — entries is never
empty here because collapsableRows is not empty. A plain else is enough.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
```diff
@@ -170,6 +170,13 @@ public ResponseEntity<String> getPoiCategories(@RequestParam String locale) {
 		}
 	}
 
+	@PostMapping(path = {"/filter-visible-tags"}, produces = "application/json")
```
alisa911: Naming: /filter-visible-tags has the same issue as the service method —
it doesn't filter. /visible-tags (a resource, not a verb) fits REST
style better.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
```diff
@@ -170,6 +170,13 @@ public ResponseEntity<String> getPoiCategories(@RequestParam String locale) {
...
+	@ResponseBody
+	public ResponseEntity<String> filterVisibleTags(@RequestBody Map<String, String> tags) {
```
alisa911: This public POST endpoint accepts an unbounded tags map with no
validation. Default Spring limits will cap the body size, but an explicit
sanity limit (e.g. max entries) would be safer.

alisa911 (commented):
Review summary for groupLocalizedTags:

The method is complex, and its name describes only one of the cases it
handles. What it really does is convert the library object into an object
for the web, so it should be named something like toVisibleTags.

Also, please split it into three cases, each in its own small method:

for (AmenityRowData row : rows) {
    VisibleTag tag = switch (row.collapsableRowType) {
        case POI_TYPE_GROUP -> toGroupTag(row);
        case PLAIN          -> toLocalizedTag(row);
        default             -> toPlainTag(row);
    };
    if (tag != null) result.add(tag);
}

https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/cb44e2e7e14554fc2f4e80877191a7b95d87bc59

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/7b9b8f3461db18e5f56a7d838f40b4dc750b00c4

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/bf90847fbd3a8b7c7fc6a02f0154a6600bf51423

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/bf90847fbd3a8b7c7fc6a02f0154a6600bf51423

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,130 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+			return types.get(0).getKeyName();
+		}
+		PoiCategory groupCategory = tagEntry.collapsableCategory;
+		for (PoiType pt : types) {
+			groupCategory = pt.getCategory();
+		}
+		return groupCategory.getKeyName();
+	}
+
+	private AmenityRowData buildWithName(AmenityRowData tagEntry, String name) {
```
alisa911: buildWithName rebuilds the row but copies only 6 of ~20 fields —
poiAdditional, collapsableCategory, collapsableExtraRow etc. are silently
dropped. It works today because toVisibleTags doesn't read them, but the
next person to use one of them after sorting gets a null. Add a
Builder.from(row) copy method to AmenityRowData in the shared code instead.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +182,130 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
...
+		if (tagEntry.collapsableRowType == AmenityRowData.CollapsableRowType.PLAIN && !Algorithms.isEmpty(tagEntry.collapsableRows)) {
+			AmenityRowData mainChild = findMainChild(tagEntry, lang);
+			key = mainChild.key;
+			value = mainChild.value;
+		}
+		PoiType pType = infoBundle.resolvePoiType(category, key, value).pType;
+		return pType != null ? pType.getKeyName() : key;
+	}
+
+	private String resolveGroupSortName(AmenityRowData tagEntry) {
```
alisa911: This "last category wins" loop over collapsablePoiTypes now exists in
three places: here, AdditionalInfoBundle.getVisibleTags and
AmenityUIHelper.buildPoiTypeGroupRowData. Move it to the shared code
(e.g. a helper in AmenityRowsBuilder) so all three stay in sync.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
```diff
@@ -170,6 +171,17 @@ public ResponseEntity<String> getPoiCategories(@RequestParam String locale) {
 		}
 	}
 
+	@PostMapping(path = {"/visible-tags"}, produces = "application/json")
+	@ResponseBody
+	public ResponseEntity<String> visibleTags(@RequestBody Map<String, String> tags,
+	                                          @RequestParam(required = false) String lang) {
+		if (tags != null && tags.size() > MAX_TAG_ENTRIES) {
+			return ResponseEntity.badRequest().body("Too many tags " + tags.size() + ", maximum is " + MAX_TAG_ENTRIES);
```
alisa911: Nit: the endpoint declares produces=application/json but the "Too many
tags" error body is plain text — return a JSON error for consistency.

Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/e0e95b08e9106096e94e1c42f5f7b519b30e0eb2

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/8da80c859d186ddd90f9810344185880e9160a46

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/8da80c859d186ddd90f9810344185880e9160a46

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
```diff
@@ -43,6 +44,8 @@
 public class SearchController {
 
 	protected static final Log LOGGER = LogFactory.getLog(SearchController.class);
+	private static final int MAX_TAG_ENTRIES = 500;
```
RZR-UA: Please justify this limit. Verify the OSM tag count for famous multilingual places such as New York, London, Paris, etc.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
```diff
@@ -180,4 +183,122 @@ public SearchUICore.SearchResultCollection addPoiCategoriesToSearchResult(
 		}
 		return resultCollection;
 	}
+
+	public List<VisibleTag> getVisibleTags(Map<String, String> tags, String lang) {
```
RZR-UA: It doesn't seem like the best class for getVisibleTags(), as that isn't connected to PoiTypes. Consider another/new class for visible-tags-related methods.

RZR-UA (approved):

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
RZR-UA: Use another name for getVisibleTags() to avoid confusion with AdditionalInfoBundle.getVisibleTags()

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/937694e409971ac1b2cafc76dc49d5ad9d04ebc8

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiTypesService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/937694e409971ac1b2cafc76dc49d5ad9d04ebc8

Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1597/commits/937694e409971ac1b2cafc76dc49d5ad9d04ebc8

RZR-UA (approved):

PR #1610, opened 2026-07-21
Fix POI search translations

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/25331

PR #1617, opened 2026-07-27
Show spatial stats for poi by category search

alisa911:


alisa911:

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
```diff
@@ -298,6 +293,32 @@ public boolean isCancelled() {
 		return response;
 	}
 
+	public Map<String, Object> getSearchStats(List<SpatialSearchContext.SpatialSearchStats> stats, long startTime,
```
vshcherb, line 296: Refactor back to single element - less code is better

alisa911: https://github.com/osmandapp/OsmAnd-tools/pull/1624

PR #1622, opened 2026-07-29
Run spatial autocomplete on a separate 3-thread pool

alisa911:


alisa911:

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
```diff
@@ -103,6 +108,20 @@ private static ThreadPoolExecutor[] createSpatialSearchExecutors() {
 		return executors;
 	}
 
+	private static ThreadPoolExecutor createAutocompleteExecutor() {
```
vshcherb, line 111: !!
1. Variables order and after methods! Static final first, class variables second by groups, services autowird 3rd

PR #1626, opened 2026-07-30
Derive per-category min zoom from frequency instead of fixed zoom gate

alisa911:


alisa911:

PR #1631, opened 2026-08-05
Refactor SpatialSearchService using shared SpatialTextSearchAPI

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/25485

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
@@ -431,6 +431,9 @@ public class SpatialSearchService {
 
 		if (spatialTextSearchAPI
 				.convertSpatialSearchResult(ssr, result, getSpatialPoiTypeSearch(), loc, locale, false) != null) {
+			if (result.objectType == ObjectType.BOUNDARY || result.objectType == ObjectType.POSTCODE) {
+				result.objectType = ObjectType.CITY; // Web does not handle POSTCODE/BOUNDARY
+			}
 			return searchResultConverter.getFeature(result, timeZone); // non-Amenity objects
 		}
 
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
@@ -75,6 +75,7 @@ public class SpatialSearchService {
 
 	private final AtomicInteger spatialSearchRoundRobin = new AtomicInteger();
 	private final Map<String, AtomicBoolean> runningSearches = new ConcurrentHashMap<>();
+	private final SpatialTextSearchAPI spatialTextSearchAPI = new SpatialTextSearchAPI(MapPoiTypes.getDefault());
 
 	// one single-thread executor per slot: requests are routed by client key, so a client's
 	// repeated searches always hit the thread whose engine cache is warmed
@@ -428,9 +429,7 @@ public class SpatialSearchService {
 		}
 		SearchResult result = new SearchResult();
 
-		SpatialTextSearchAPI api = new SpatialTextSearchAPI(MapPoiTypes.getDefault());
-
-		if (api.convertSpatialSearchResult(ssr, result, poiSearch, loc, locale, false) != null) {
+		if (spatialTextSearchAPI.convertSpatialSearchResult(ssr, result, poiSearch, loc, locale, false) != null) {
 			return searchResultConverter.getFeature(result, timeZone); // non-Amenity objects
 		}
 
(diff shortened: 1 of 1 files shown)


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
```diff
@@ -431,60 +427,13 @@ private Feature getSpatialFeature(LatLon loc, List<MapObject> objs, String local
...
 
-	private Street getSpatialStreet(List<MapObject> objs) {
-		for (MapObject obj : objs) {
-			if (obj instanceof Street street) {
-				return street;
-			}
-		}
-		return null;
-	}
+		SpatialTextSearchAPI api = new SpatialTextSearchAPI(MapPoiTypes.getDefault());
```
alisa911: Please move this to a service field or create once per request. It's allocated for every search result and its constructor also creates a new SpatialTextSearch.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
```diff
@@ -431,60 +427,13 @@ private Feature getSpatialFeature(LatLon loc, List<MapObject> objs, String local
...
 
-	private City getSpatialCity(List<MapObject> objs) {
-		for (MapObject obj : objs) {
-			if (obj instanceof City city) {
-				return city;
-			}
-			if (obj instanceof Street street && street.getCity() != null) {
-				return street.getCity();
-			}
+		if (api.convertSpatialSearchResult(ssr, result, poiSearch, loc, locale, false) != null) {
```
alisa911: Better to pass getSpatialPoiTypeSearch() instead of the raw poiSearch field to avoid a possible null and keep lazy init in one place.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
```diff
@@ -431,60 +427,13 @@ private Feature getSpatialFeature(LatLon loc, List<MapObject> objs, String local
...
+			return searchResultConverter.getFeature(result, timeZone); // non-Amenity objects
```
alisa911, line 437: Note this changes the response: City results now come as VILLAGE/POSTCODE/BOUNDARY instead of CITY, and web doesn't handle POSTCODE/BOUNDARY — was this checked on the frontend?

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
alisa911: Multi-street results now get STREET_INTERSECTION, and the converter renders them as "Street - City" (or "Street - null" if no city) — should probably stay STREET or fix the converter.

alisa911: old code null-checked the main object, but getMainObject() inside the API can return null while getObjects() is non-empty

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
RZR-UA: e08ea7add

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
RZR-UA: 39e580c32

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
RZR-UA: ca2dd0fa6

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
RZR-UA: Wtf is "converter renders"? Now it is the same result in Android and Web. STREET_INTERSECTION shown with special type and special icon and the Street (1st of 2) shown in the same way.

RZR-UA: getMainObject() should never return null if getObjects() is non-empty, but I'll add additional protection

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/SpatialSearchService.java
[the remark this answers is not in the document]
RZR-UA: fixed

PR #1641, opened 2026-08-20
Bbox poi spatial

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/25331
_4.2 Search in large parks, neighborhood same as in boundaries (index bbox POI), residential_

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexPoiCreator.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexPoiCreator.java
@@ -6,9 +6,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.RandomAccessFile;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.sql.*;
 import java.util.*;
 import java.util.Map.Entry;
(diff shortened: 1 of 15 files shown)


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexPoiCreator.java
```diff
@@ -203,7 +203,8 @@ void iterateEntityInternal(Entity e, OsmDbAccessorContext ctx, IndexCreationCont
 			
 			List<LatLon> relationCenters = Collections.singletonList(null); // [null] means single amenity point
 			StringBuilder memberIds = new StringBuilder();
-			relationCenters = collectRelationCenters(e, ctx, tags, relationCenters, memberIds);
+			QuadRect indexedBbox = OsmMapUtils.requireIndexBbox(tags) ? new QuadRect() : null;
```
vshcherb, line 216: MINIMUM POLYGON SIZE is needed. MIN size needs to be investigeted

ivanPyrohivskyi: Added MIN_BBOX_INDEX_SIZE_KM
LATITUDE_IN_1_KM - is always constant, longitude is a changing value

vshcherb:
https://github.com/osmandapp/OsmAnd-tools/pull/1647

vshcherb:
Unnecessary commits deleted

PR #1644, opened 2026-08-22
Add authenticated empty-trash endpoint

alex-dev-neo:
## Summary

- expose authenticated `POST /userdata/empty-trash` for mobile clients
- accept the existing `{name,type,updatetime}` batch payload
- validate the complete batch before deleting anything
- purge every historical version of each validated Trash item
- reuse the validation and deletion logic in `deleteFileAllVersions()`

## Context

Mobile clients previously deleted only the tombstone and one preceding version. Older versions could remain and reappear after synchronization.

Related to osmandapp/OsmAnd#24228.

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
@@ -50,6 +50,9 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.web.multipart.MultipartFile;
 
 import com.google.common.cache.Cache;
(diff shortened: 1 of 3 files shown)


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -800,8 +800,24 @@ public boolean checkIfRestoredVersionExists(String name, String type, Long updat
     
     @Transactional
     public ResponseEntity<String> emptyTrash(List<MapApiController.FileData> files, CloudUserDevicesRepository.CloudUserDevice dev) {
```
alisa911: This logic repeats the same validation and delete flow already used in deleteFileAllVersions(). It would be cleaner to reuse one shared helper instead of duplicating the logic.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -923,20 +939,32 @@ public ResponseEntity<String> deleteFileVersion(Long updatetime, int userid, Str
...
         List<UserFile> files = filesRepository.findAllByUseridAndNameAndTypeOrderByUpdatetimeDesc(userid, fileName, fileType);
+        String error = validateFileVersionsDeletion(files, updatetime, isTrash);
+        if (error != null) {
+            return ResponseEntity.badRequest().body(error);
+        }
+        deleteFileVersions(files);
+        return ok();
+    }
+
+    private String validateFileVersionsDeletion(List<UserFile> files, Long updatetime, boolean isTrash) {
```
alisa911: The naming here is a bit unclear. A clearer name like validateVersionsForDeletion or validateFileVersionsBeforeDelete would be easier to read and understand.

alisa911 (commented):
This removes the versions, but not the related info files. We should probably handle that here too so the cleanup stays consistent.

PR #1645, opened 2026-08-24
Add ferry roads generation

nnngrach:
[issue](https://github.com/osmandapp/OsmAnd/issues/17773)
[related android request](https://github.com/osmandapp/OsmAnd/pull/25725)

---

**Gullmarsleden**

58.30263 11.50282
58.29961 11.53579
[https://www.openstreetmap.org/relation/7841117](https://www.openstreetmap.org/relation/7841117#map=15/58.30284/11.52026&layers=Y)

Before
Screenshot 2026-08-24 at 14 55 54   Screenshot 2026-08-24 at 14 56 05

After (tested on new generated small obf)
Screenshot 2026-08-24 at 14 58 53 Screenshot 2026-08-24 at 14 59 43

Screenshot 2026-08-25 at 02 46 23

---

**Sandbanks Ferry**

50.67782 -1.95123
50.68383 -1.94831
https://www.openstreetmap.org/way/147985346

Before
Screenshot 2026-08-24 at 19 21 23      Screenshot 2026-08-24 at 19 21 59

After
Screenshot 2026-08-24 at 19 23 06      Screenshot 2026-08-24 at 19 23 20

Screenshot 2026-08-25 at 02 52 58

---

**Nordoleden**

57.75831 11.61556
57.77241 11.61966
https://www.openstreetmap.org/way/16794766

Before
Screenshot 2026-08-24 at 20 09 36    Screenshot 2026-08-24 at 20 09 52

After
Screenshot 2026-08-25 at 01 15 48   Screenshot 2026-08-25 at 01 15 57

Screenshot 2026-08-25 at 02 49 43

---

[ferry obf files for test.zip](https://github.com/user-attachments/files/31394008/ferry.obf.files.for.test.zip)

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexCreator.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexCreator.java
@@ -756,8 +756,9 @@ public class IndexCreator {
 			public void iterateEntity(Entity e, OsmDbAccessorContext ctx) throws SQLException {
                 Relation r = (Relation) e;
                 if (IndexFerryHelper.hasFerryTags(r)) {
-                    iterateMainEntity(e, ctx, icc); //create ferry road from real good-tagged relation.
-					IndexFerryHelper.removeDuplicatedFerryWays(r, collectedFerryWays); //remove way if we have relation
+					//if we have real good-tagged relation, then remove way and create road from relation.
+                    iterateMainEntity(e, ctx, icc); 
+					IndexFerryHelper.removeDuplicatedFerryWays(r, collectedFerryWays);
                 } else {
                     iterateMainEntity(e, ctx, icc);
                 }
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexCreator.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexCreator.java
@@ -8,10 +8,12 @@ import net.osmand.impl.ConsoleProgressImplementation;
 import net.osmand.obf.preparation.OsmDbAccessor.OsmDbVisitor;
 import net.osmand.osm.MapPoiTypes;
 import net.osmand.osm.MapRenderingTypesEncoder;
-import net.osmand.osm.edit.OSMSettings.OSMTagKey;
-import net.osmand.osm.edit.*;
+import net.osmand.osm.edit.Entity;
 import net.osmand.osm.edit.Entity.EntityId;
 import net.osmand.osm.edit.Entity.EntityType;
+import net.osmand.osm.edit.Node;
+import net.osmand.osm.edit.Relation;
+import net.osmand.osm.edit.Way;
 import net.osmand.osm.io.IOsmStorageFilter;
 import net.osmand.osm.io.OsmBaseStorage;
 import net.osmand.osm.io.OsmBaseStoragePbf;
@@ -27,7 +29,9 @@ import javax.imageio.ImageReader;
 import java.io.*;
 import java.sql.Connection;
 import java.sql.SQLException;
-import java.util.*;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.zip.GZIPInputStream;
 
 /**
@@ -724,7 +728,8 @@ public class IndexCreator {
 
 	private void iterateMainEntities(OsmDbAccessor accessor, IProgress progress, IndexCreationContext icc)
 			throws SQLException, InterruptedException {
-        Map<String, Way> foundFerryWays = new LinkedHashMap<String, Way>();
+        Map<String, Way> foundedFerryWays = new LinkedHashMap<String, Way>();
+		
 		setGeneralProgress(progress, "[50 / 100]");
 		progress.startTask(settings.getString("IndexCreator.PROCESS_OSM_NODES"), accessor.getAllNodes());
 		accessor.iterateOverEntities(progress, EntityType.NODE, new OsmDbVisitor() {
@@ -739,7 +744,7 @@ public class IndexCreator {
 			@Override
 			public void iterateEntity(Entity e, OsmDbAccessorContext ctx) throws SQLException {
 				Way w = (Way) e;
-                saveFoundedFerry(w, foundFerryWays);
+				IndexFerryHelper.saveFoundedFerryWay(w, foundedFerryWays);
 				propagateToNodes.calculateBorderPoints(w);
 				iterateMainEntity(e, ctx, icc);
 			}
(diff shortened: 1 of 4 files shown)


java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexFerryHelper.java
```diff
@@ -0,0 +1,96 @@
+package net.osmand.obf.preparation;
```
vshcherb, line 1: Naming ... not Helper for sure IndexFerry...

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexCreator.java
```diff
@@ -749,9 +757,23 @@ public void iterateEntity(Entity e, OsmDbAccessorContext ctx) throws SQLExceptio
 		accessor.iterateOverEntities(progress, EntityType.RELATION, new OsmDbVisitor() {
 			@Override
 			public void iterateEntity(Entity e, OsmDbAccessorContext ctx) throws SQLException {
-				iterateMainEntity(e, ctx, icc);
+                Relation r = (Relation) e;
```
vshcherb, line 760: Bad place for code if it's related to transport should be in trasport creator or subclass delegated from it

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexCreator.java
```diff
@@ -749,9 +757,23 @@ public void iterateEntity(Entity e, OsmDbAccessorContext ctx) throws SQLExceptio
...
+					//if we have real good-tagged relation, then remove way and create road from relation.
+                    iterateMainEntity(e, ctx, icc); 
+					IndexFerryHelper.removeDuplicatedFerryWays(r, collectedFerryWays);
+                } else {
+                    iterateMainEntity(e, ctx, icc);
+                }
 			}
 		});
+
+        for (Way way : collectedFerryWays.values()) {
```
vshcherb, line 772: Looks strange as well

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexFerryHelper.java
```diff
@@ -0,0 +1,96 @@
...
+    public static void removeDuplicatedFerryWays(Relation rel, Map<String, Way> collectedFerryWays) {
+        if (hasFerryTags(rel)) {
+            long startStopId = rel.getMembers().get(0).getEntity().getId();
+            long endStopId = rel.getMembers().get(1).getEntity().getId();
+            collectedFerryWays.remove(startStopId + " " + endStopId);
+            collectedFerryWays.remove(endStopId + " " + startStopId);
+        }
+    }
+
+    public static Relation createSyntheticFerryRelation(Way way, Set<Long> usedRelationIds) {
```
vshcherb, line 64: Why synthesize a Relation here instead of feeding the Way into IndexTransportCreator directly? I assume it's because most ferry crossings in OSM are tagged only on the way itself (`ferry=*`/`route=ferry`), without a proper `type=route;route=ferry` relation wrapping it — and the whole PT pipeline (processTransportRelationV1/V2) is relation-based, so a bare way would never produce a TransportRoute otherwise. Worth stating that rationale explicitly in a comment here.

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexFerryHelper.java
```diff
@@ -0,0 +1,96 @@
...
+        //create relation based on ferry-way dato. it will be used for generation forward and backward public transport routes.
+        long syntheticRelationId = generateSyntheticRelationId(way, usedRelationIds);
+        Relation syntethicRelation = new Relation(syntheticRelationId);
```
vshcherb, line 67: This relation gets a fake id that never existed in OSM, but nothing marks it as synthetic. ObfConstants.isOsmUrlAvailable()/getOsmUrlForId() only check id > 0, so the UI will happily build an openstreetmap.org/relation/<id> link for it (route info / 'open in OSM'), pointing to a relation that doesn't correspond to this object (or a real, unrelated one once IDs grow past 2^26). Should we add an isSyntheticFerryRelationId(long id) check to ObfConstants (mirroring isIdFromPropagatedNode/isIdFromSplit) and use it to suppress the 'open in OSM' link for these, same as for any other synthetic/transport map object?

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexFerryHelper.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+    
+    public static boolean hasFerryTags(Entity e) {
+        // "ferry=*" or "route=ferry"
+        return e.getTag(FERRY) != null || (e.getTag(ROUTE) != null && e.getTag(ROUTE).equals(FERRY.getValue()));
+    }
+
+    public static void collectFoundedFerryWays(Way way, Map<String, Way> collectedFerryWays) {
+        if (hasFerryTags(way)) {
+            String key = way.getFirstNodeId() + " " + way.getLastNodeId();
```
vshcherb, line 50: Key collision: two distinct ferry lines sharing the same two terminal nodes (e.g. two operators/parallel routes between the same ports) will collide on this key and silently overwrite each other in the map — one of them is dropped from generation with no warning.

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexFerryHelper.java
```diff
@@ -0,0 +1,96 @@
...
+            collectedFerryWays.put(key, way);
+        }
+    }
+
+    public static void removeDuplicatedFerryWays(Relation rel, Map<String, Way> collectedFerryWays) {
+        if (hasFerryTags(rel)) {
+            long startStopId = rel.getMembers().get(0).getEntity().getId();
```
vshcherb, line 57: This assumes members[0]/members[1] are the start/end stop nodes. Real (non-synthetic) OSM route relations don't guarantee member order — a relation with a way member first, or with <2 members, will either silently fail to dedupe (get the wrong id here) or throw IndexOutOfBoundsException on get(1).

java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/obf/preparation/IndexFerryHelper.java
```diff
@@ -0,0 +1,96 @@
...
+            long endStopId = rel.getMembers().get(1).getEntity().getId();
+            collectedFerryWays.remove(startStopId + " " + endStopId);
```
vshcherb, line 59: This only dedupes when the whole ferry route is a single way (its own first/last node matches the relation's start/end stop). If a real relation's ferry route is split across multiple way segments, none of the fragments will match this key, so they stay in collectedFerryWays and each gets its own synthetic relation later — duplicate/overlapping routes alongside the real one.

PR #1661, opened 2026-08-31
Read the obf indexes through indexes.cache in the working folder

vshcherb:
Fixes the `WARN: Native File not initialized from cache: /home/tileserver/maps/....obf` flood in the build log.

## Cause

`initAllMaps` opened every map with a plain `initMapFile`, so the native library never got an index cache: it warned once per map and re-read the index of all of them on every run.

## Fix

The tester now builds `indexes.cache` and hands it to `initCacheMapFile` before opening the maps. The cache goes into the **working folder** — `-cache.dir`, default `user.dir`, so on Jenkins it lands in `$WORKSPACE` and survives between builds — and never into the maps folder, which on a tile server is shared and may be read only.

Measured with 126 maps locally:

| | init | warnings |
|---|---|---|
| cold, no cache | 11.6 s | 0 |
| second run, cache reused | **2.4 s** | 0 |

The gain grows with the number of maps; the builder loads 1078.

## Two supporting changes in NativeJavaRendering

- `initIndexesCache` got an overload that takes the cache file explicitly. The existing three argument method still writes into the maps folder, so no other caller changes behaviour.
- **A broken obf no longer aborts the whole folder.** `Uruguay_southamerica_2.obf` in a local maps folder is corrupt, and going through the index cache turned that into a fatal `IOException: Corrupt file, it should have ended as it starts with version` that killed the entire run before a single tile was rendered. It is now logged and skipped, and also kept out of the list of maps to open. Worth a look during review — it changes shared behaviour, but one bad file killing a 1078 map scan seems clearly wrong.

```
SEVERE: Skipping map .../Uruguay_southamerica_2.obf: Corrupt file, ...
Indexes cache : .../indexes.cache (created)
Initialized 126 maps from /Users/.../maps
```

https://claude.ai/code/session_012ewhEQGVv79BpPExDNXmxc

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/NativeJavaRendering.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/NativeJavaRendering.java
@@ -538,8 +538,16 @@ public class NativeJavaRendering extends NativeLibrary {
 	}
 
 	public Map<String, FileIndex> initIndexesCache(File dir, List<File> filesToUse, boolean filterDuplicates) throws IOException {
+		return initIndexesCache(dir, new File(dir, INDEXES_CACHE), filesToUse, filterDuplicates);
+	}
+
+	/**
+	 * Same, but the cache is kept in {@code cacheFile} instead of the maps folder itself, for the
+	 * callers that must not write into a shared or read only maps folder.
+	 */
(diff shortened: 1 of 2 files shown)


vshcherb:
Superseded: the indexes cache is fixed to the folder the job runs in (the Jenkins workspace), with no option to configure it, and the change is handed over as a working tree diff instead of a PR.

PR #1664, opened 2026-09-02
Server: return HH alternative routes from /routing/route

vshcherb:
Returns the alternative routes that HH routing finds from `/routing/route`, so the map can draw them. Needs osmandapp/OsmAnd#25805 for the router itself. Ref osmandapp/OsmAnd-Issues#2855.

## API

New request parameter `alternatives`, 0 by default — nothing changes for callers that do not ask. When set, HH routing is asked for that many alternatives and each one is appended to the FeatureCollection as its own LineString after the main route:

| property | meaning |
|---|---|
| `alternative` | 1-based index; marks the route as not the one being shown |
| `overall.distance` / `overall.time` / `overall.routingTime` | the same block the main route carries |
| `diffElevationUp` / `diffElevationDown` | as for the main route |
| `deltaTime` | overhead against the main route, percent |

Alternatives get their elevation from the same `getElevationsBySegments()` the main route goes through, and the same `overall` shape, so a client can put an alternative in place of the route it is showing and have the summary and the elevation graph keep working. That is what osmandapp/web#1975 does on click.

The HH config can be shared with the routing cache, so `CALC_ALTERNATIVES` is set for the request only and put back before the context is unlocked.

## Testing

Local server over `Ukraine_kyiv_europe_2.obf`, Podil → Teremky (`50.46904,30.51638` → `50.37401,30.44741`): two alternatives, 16.2 km (+0.5%) and 27.5 km (+9.4%). Checked both directly and through the map's dev proxy.

Worth knowing when comparing against the unit tests: this pair has two nearly tied optima — 25.5 km at 42 km/h round the west and 16.2 km at 23 km/h through the centre, 1881 against 1877 of cost. Which one comes back as the main route depends on the search order, and the other one is offered as an alternative at +0.5%. Nothing is wrong when they swap places between two setups.

---

## AI disclaimer

Written with Claude Code (Claude Opus 5), driven by @vshcherb. Summary of the requests it worked from, in order:

1. Analyse how to do alternative routes in OsmAnd; `HHRoutePlanner` is the main class; read the fast-routing blog post to see how HH works. What is wanted is *interesting, correct* alternatives — maximally different geometry at a similar time, not a detour around a pole for +5%. 1–2 alternatives always, and it has to be fast. The existing ideas are tied to opaque heuristics in the code; wanted an understandable one, roughly Google-like but not a black box, so it can be tuned on user feedback later. Run tests on the local maps, and decide what to compare against and which criteria to use.
2. Turn it into code, and draw the alternatives so they can be tested with HH.
3. Not the Android app — the thing to run is OsmExtractionUI (OsmAndMapCreator, tools project), `MapRouterLayer`.
4. Both routes have the same colour — give the second one a different colour. Which one is the optimal?
5. Too many labels on alt 1 / alt 2, they flood the map. Use green for alt 1 and magenta for alt 2. Reverse the draw order — alt 2 first, then alt 1, then the main route, so the main route is always fully visible on top.
6. Open pull requests from the tools and android branches.

All numbers in this description are measurements from those runs, not estimates.

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
@@ -1098,11 +1098,39 @@ public class OsmAndMapsService {
 					hhConfig.ALT_MAX_COUNT = keepAlternatives;
 				}
 			}
+			if (alternatives > 0) {
+				String reason = alternativesUnsupportedBy(rp, ctx, router, intermediates);
+				if (reason != null) {
+					// they would silently come back empty otherwise, which reads as a broken feature
+					props.put("alternativesNotSupported", reason);
+				}
+			}
 		} finally {
 			unlockReaders(usedMapList);
 			unlockCacheRoutingContext(ctx);
 		}
 		return routeRes;
+	}
+
+	/**
+	 * Only the Java HH planner produces alternatives, and only for a plain start -> end route.
+	 * Returns what stands in the way, or null when alternatives can be expected.
+	 */
+	private String alternativesUnsupportedBy(RouteParameters rp, RoutingContext ctx,
+			RoutePlannerFrontEnd router, List<LatLon> intermediates) {
+		if (rp.onlineRouting != null) {
+			return "online routing";
+		}
+		if (ctx.nativeLib != null) {
+			return "native (C++) routing";
+		}
+		if (!router.isHHRoutingConfigured()) {
+			return "HH routing is off";
+		}
+		if (intermediates != null && !intermediates.isEmpty()) {
+			return "intermediate points";
+		}
+		return null;
 	}
 
 	private static long getLocalTimeMillisByLatLon(double lat, double lon) {
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
@@ -1024,6 +1024,17 @@ public class OsmAndMapsService {
 	public List<RouteSegmentResult> routing(boolean disableOldRouting, String routeMode, Map<String, Object> props,
 	                                        LatLon start, LatLon end, List<LatLon> intermediates, List<String> avoidRoadsIds,
 	                                        RouteCalculationProgress progress) throws IOException, InterruptedException {
+		return routing(disableOldRouting, routeMode, props, start, end, intermediates, avoidRoadsIds, progress, 0, null);
+	}
+
+	/**
+	 * @param alternatives how many alternative routes to ask HH routing for (0 - none)
+	 * @param alternativesOut filled with the alternatives that were found, may be null
+	 */
+	public List<RouteSegmentResult> routing(boolean disableOldRouting, String routeMode, Map<String, Object> props,
+	                                        LatLon start, LatLon end, List<LatLon> intermediates, List<String> avoidRoadsIds,
+	                                        RouteCalculationProgress progress, int alternatives,
+	                                        List<List<RouteSegmentResult>> alternativesOut) throws IOException, InterruptedException {
 		String profile = routeMode.split("\\,")[0];
 		QuadRect points = points(intermediates, start, end);
 		RoutePlannerFrontEnd router = new RoutePlannerFrontEnd();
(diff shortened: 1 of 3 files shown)


java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/RoutingController.java
```diff
@@ -290,7 +290,8 @@ public ResponseEntity<String> geocoding(@RequestParam double lat, @RequestParam
 	public ResponseEntity<String> routing(HttpSession session,
 			@RequestParam String[] points, @RequestParam(defaultValue = "car") String routeMode,
 	                                 @RequestParam(required = false) String[] avoidRoads,
-	                                 @RequestParam(defaultValue = "production") String limits) throws IOException {
+	                                 @RequestParam(defaultValue = "production") String limits,
+	                                 @RequestParam(defaultValue = "0") int alternatives) throws IOException {
```
alisa911, line 296: The parameter is client controlled and unbounded. Clamp it to a small max like 2 or 3 before passing to routing

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/OsmAndMapsService.java
```diff
@@ -1061,13 +1072,31 @@ public List<RouteSegmentResult> routing(boolean disableOldRouting, String routeM
...
+			int keepAlternatives = hhConfig == null ? 0 : hhConfig.ALT_MAX_COUNT;
+			boolean keepCalcAlternatives = hhConfig != null && hhConfig.CALC_ALTERNATIVES;
+			if (hhConfig != null) {
+				hhConfig.calcAlternative(alternatives);
+			}
+			try {
+				if (rp.onlineRouting != null) {
+					routeRes = onlineRouting(rp, ctx, router, props, start, end, intermediates);
+				} else {
+					RouteCalcResult rc = ctx.nativeLib != null ? runRoutingSync(start, end, intermediates, router, ctx)
```
alisa911, line 1087: Alternatives silently come back empty for hh_cpp, native and online routing since only the Java HH planner fills them. Worth a note in the API or a warning property in the response.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/RoutingService.java
```diff
@@ -211,6 +211,64 @@ public void interpolateEmptyElevationSegments(List<LatLonEle> points) {
...
+            overall.put("routingTime", routingTime(alt));
+            f.properties.put("overall", overall);
+            List<Double> eleDiff = calculateElevationDiffs(points);
+            if (eleDiff.size() > 1 && !Double.isNaN(eleDiff.get(0)) && !Double.isNaN(eleDiff.get(1))) {
+                f.properties.put("diffElevationUp", eleDiff.get(0));
+                f.properties.put("diffElevationDown", eleDiff.get(1));
+            }
+            if (mainRoutingTime > 0) {
+                double delta = 100 * (routingTime(alt) / mainRoutingTime - 1);
+                f.properties.put("deltaTime", Math.round(delta * 10) / 10.0);
```
alisa911: deltaTime is computed from routing cost, not time, and the baseline differs from the main route overall.routingTime which is ctx.routingTime. Either compute both sides the same way and rename, or drop the property since the client does not use it.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/RoutingService.java
```diff
@@ -211,6 +211,64 @@ public void interpolateEmptyElevationSegments(List<LatLonEle> points) {
...
+            List<RouteSegmentResult> alt = alternatives.get(i);
+            double distance = 0, time = 0;
+            for (RouteSegmentResult r : alt) {
+                distance += r.getDistance();
+                time += r.getSegmentTime();
+            }
+            // same elevation treatment as the main route, so the client can show its graph too
+            List<LatLonEle> points = getElevationsBySegments(new ArrayList<>(), new ArrayList<>(), alt);
+            interpolateEmptyElevationSegments(points);
+            if (points.size() < 2) {
```
alisa911: Skipping an alternative leaves a gap in the alternative numbering. Number the features by a separate counter.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/RoutingService.java
```diff
@@ -211,6 +211,64 @@ public void interpolateEmptyElevationSegments(List<LatLonEle> points) {
...
+        if (alternatives == null || alternatives.isEmpty()) {
+            return features;
+        }
+        double mainRoutingTime = routingTime(mainRoute);
+        for (int i = 0; i < alternatives.size(); i++) {
+            List<RouteSegmentResult> alt = alternatives.get(i);
+            double distance = 0, time = 0;
+            for (RouteSegmentResult r : alt) {
+                distance += r.getDistance();
+                time += r.getSegmentTime();
```
alisa911, line 244: Turn description points are dropped for alternatives, so the client cannot show turns after swapping. Consider returning them tagged with the alternative index.

PR #1665, opened 2026-09-02
Encode osmand_region_ref with ref:USPS / ISO3166-2 (LLM)

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/25485

## AI disclaimer

Produced with Codex (GPT-5), including the added and revised tests.

Prompts used (summarised):

1. Investigate issue #25485 and identify the first bug involving region references and regions.ocbf.
2. Analyse OCBF contents and generation for FR-55, numeric 55, and Pennsylvania PA.
3. Generate osmand_region_ref from ref:USPS, falling back to ISO3166-2, while preserving the original ref.
4. Update the client, constants, tests, and region matching to use the new field; remove the one-character bypass.
5. Generate a fresh local OCBF, inspect its dump, and run the relevant tests, excluding OsmandRegionsTest until explicitly requested.
6. Refine matching so PA remains available to address search while 55 does not match FR-55; use a separate regionSearchRef and exact ref comparison.
7. Update the existing OsmandRegionsTest and restore shared matcher reuse.

Decided by the agent, not requested explicitly: registered osmand_region_ref as a non-localized OCBF text field, added regionSearchRef to WorldRegion, and used a temporary local OCBF generation and test workflow for validation.

### DIFF
--- a/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/util/CountryOcbfGeneration.java
+++ b/java-tools/OsmAndMapCreatorUtilities/src/main/java/net/osmand/util/CountryOcbfGeneration.java
@@ -26,6 +26,7 @@ import java.util.TreeMap;
 import net.osmand.PlatformUtil;
 import net.osmand.binary.MapZooms;
 import net.osmand.impl.ConsoleProgressImplementation;
+import net.osmand.map.OsmandRegions;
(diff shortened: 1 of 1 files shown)


RZR-UA:
Closing because the proposed separate osmand_region_ref field will not be used.

PR #1680, opened 2026-09-08
Follow the routing classes that moved to OsmAnd-shared

alex-dev-neo:
`java-tools` builds `OsmAnd-java` and `OsmAnd-shared` from source — `settings.gradle` points at `../../android/` — so the ongoing move of routing classes into `OsmAnd-shared` breaks compilation here.

Already broken on master once osmandapp/OsmAnd#25901 and osmandapp/OsmAnd#25903 land:

- `MapRouterLayer` uses `TurnType` through `import net.osmand.router.*`
- `SearchResultConverter` uses `net.osmand.util.OpeningHoursParser`

Broken by osmandapp/OsmAnd#25904 as well: twelve files that reach `RouteTypeRule` through `BinaryMapRouteReaderAdapter`.

Most of it is imports. Two call sites needed more:

- **`SearchResultConverter`** built a `java.util.Calendar` in the client's time zone to evaluate opening hours. The shared parser takes an `OpeningHoursTime`, which gains a factory for an explicit zone in osmandapp/OsmAnd#25904. An unknown zone id falls back to UTC there, matching what `TimeZone.getTimeZone` did with GMT.
- **`RoutingService`** called `RouteTypeRule.writeToBundle`, which went away with the `StringExternalizable` interface. It writes the two keys itself now, the same way `RouteExporter` does.

**Merge order:** osmandapp/OsmAnd#25900 → #25901 (with osmandapp/OsmAnd-core-legacy#350) → #25903 → #25904, then this.

Verified with `./gradlew compileJava compileTestJava` against the `shared-routing-4-route-type-rule` branch — all four projects compile.


alex-dev-neo:

PR #1681, opened 2026-09-08
Follow RouteDataObject and the route regions to OsmAnd-shared

alex-dev-neo:
Keeps `java-tools` compiling against osmandapp/OsmAnd#25906, which moves `RouteDataObject`, `RouteRegion`, `RouteSubregion` and `BinaryIndexPart` into the `OsmAnd-shared` KMP module. Stacked on #1680.

29 files, almost all of it imports:

- `net.osmand.binary.RouteDataObject` → `net.osmand.shared.routing.RouteDataObject`
- `BinaryMapRouteReaderAdapter.RouteRegion` / `.RouteSubregion` → `net.osmand.shared.routing.RouteRegion` / `.RouteSubregion`, no longer nested
- `net.osmand.binary.BinaryIndexPart` → `net.osmand.shared.binary.BinaryIndexPart`

Three files also follow `RouteDataObject.names` from `gnu.trove.map.hash.TIntObjectHashMap` to `KTIntObjectMap`: `BinaryInspector`, `IndexRouteCreator` and `BinaryMapIndexTestReader`. `KTIntObjectMap` keeps the trove shaped api — `keys()`, `get`, `put`, `size()` and the `hasNext`/`advance`/`key`/`value` iterator — so only the declared types change.

**Merge order:** osmandapp/OsmAnd-core-legacy#351 → OsmAndCore snapshot rebuild → osmandapp/OsmAnd#25906 → this.

Verified with `./gradlew compileJava compileTestJava`: all four projects compile against the `shared-routing-6-route-data-object` branch.

PR #1682, opened 2026-09-08
Follow the route bundles to OsmAnd-shared

alex-dev-neo:
Keeps `java-tools` compiling against osmandapp/OsmAnd#25908, which moves `RouteDataBundle`, `RouteDataResources` and `StringExternalizable` into `net.osmand.shared`. Stacked on #1681.

One file, `RoutingService`:

- the bundles are built on the shared `StringBundle` now, so `generateRouteSegments` hands them straight to `GpxUtilities.RouteSegment.fromStringBundle`. `toKotlinStringBundle` existed only to bridge the two `StringBundle` implementations and goes away.
- `RouteDataResources` holds `KLocation`, so the locations are converted once with the new `Location.toShared`.

**Merge order:** osmandapp/OsmAnd-core-legacy#351 → OsmAndCore snapshot rebuild → the OsmAnd stack through #25908 → #1680, #1681, this.

Verified with `./gradlew compileJava compileTestJava` against the `shared-routing-7-route-bundles` branch.

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/RoutingService.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/RoutingService.java
@@ -18,7 +18,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 import net.osmand.Location;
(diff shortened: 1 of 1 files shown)


alex-dev-neo:

PR #1683, opened 2026-09-08
Follow RouteSegmentResult to OsmAnd-shared

alex-dev-neo:
Keeps `java-tools` compiling against osmandapp/OsmAnd#25911, which moves `RouteSegmentResult` into the `OsmAnd-shared` KMP module. Stacked on #1682.

Eight files change an import. Two also follow the point accessors: `getStartPoint`, `getEndPoint` and `getPoint` answer with a `KLatLon` now, because the shared module has no `LatLon`. `MapRouterLayer` and `RoutingService` convert with the new `LatLon.of`, or measure distances with `KMapUtils` directly where the result was only fed to `MapUtils.getDistance`.

**Merge order:** osmandapp/OsmAnd-core-legacy#350, #351, #352 → OsmAndCore snapshot rebuild → the OsmAnd stack through #25909 → #1680, #1681, #1682, this.

Verified with `./gradlew compileJava compileTestJava` against the `shared-routing-8-route-segment-result` branch.


alex-dev-neo:

PR #1684, opened 2026-09-08
Follow VehicleRouter to OsmAnd-shared

alex-dev-neo:
Keeps `java-tools` compiling against osmandapp/OsmAnd#25912, which moves `VehicleRouter` into the `OsmAnd-shared` KMP module. Stacked on #1683.

One import, in `ImproveRoadConnectivity`, which only names the type — it calls `acceptLine`, `isOneWay` and the speed methods, none of which changed.

**Merge order:** osmandapp/OsmAnd-core-legacy#350, #351, #352 → OsmAndCore snapshot rebuild → the OsmAnd stack through #25912 → #1680 to #1683, this.

Verified with `./gradlew compileJava compileTestJava` against the `shared-routing-9a-vehicle-router` branch.


alex-dev-neo:

PR #353, opened 2020-05-16
Route merge

vshcherb:

PR #1685, opened 2026-09-08
Follow GeneralRouter to OsmAnd-shared

alex-dev-neo:
Keeps `java-tools` compiling against osmandapp/OsmAnd#25913, which moves `GeneralRouter` and its nested classes into the `OsmAnd-shared` KMP module. Stacked on #1684.

Imports only, in eight files: the class itself, plus the nested `RouteDataObjectAttribute`, `RouteAttributeContext`, `RoutingParameter` and `RoutingParameterType` where they were imported by name. The api they use — `getObjContext`, `registerBooleanParameter`, `registerNumericParameter`, `getParameters`, the evaluate methods — is unchanged.

**Merge order:** osmandapp/OsmAnd-core-legacy#350 to #353 → OsmAndCore snapshot rebuild → the OsmAnd stack through #25913 → #1680 to #1684, this.

Verified with `./gradlew compileJava compileTestJava` against the `shared-routing-9b-general-router` branch.


alex-dev-neo:

PR #1686, opened 2026-09-08
Follow RouteCalculationProgress to OsmAnd-shared

alex-dev-neo:
Stacked on #1685, review that one first. Follows osmandapp/OsmAnd#25917.

`RouteCalculationProgress` moved from `net.osmand.router` to `net.osmand.shared.routing`, together with `FastRoutingState`. Imports only, in seven files across `OsmAndMapCreator`, `OsmAndMapCreatorUtilities` and `OsmAndServer`.

All four java-tools projects compile.

Merges after the OsmAnd side.


alex-dev-neo:

PR #1687, opened 2026-09-08
Follow direction points to OsmAnd-shared

alex-dev-neo:
Stacked on #1686, review that one first. Follows osmandapp/OsmAnd#25919.

`RoutingConfiguration.Builder.setDirectionPoints` takes a `KQuadTree<DirectionPoint>` now instead of a `QuadTree` of OSM nodes, so `MapRouterLayer` builds direction points from the GeoJSON it loads rather than nodes.

The map viewer draws `Entity`, so the points are still shown through a node carrying their tags.

All four java-tools projects compile.

Merges after the OsmAnd side.


alex-dev-neo:

PR #1688, opened 2026-09-08
Follow RoutingConfiguration to OsmAnd-shared

alex-dev-neo:
Stacked on #1687, review that one first. Follows osmandapp/OsmAnd#25920.

`RoutingConfiguration` moved from `net.osmand.router` to `net.osmand.shared.routing`. Imports in twelve files, and one parse call.

`parseFromInputStream(InputStream)` is gone — it takes an okio `Source` now, with a path overload for Java callers — so `DataExtractionSettings` hands it the path it already had. The two checked exceptions that call site caught went with the stream, so it catches what the shared parser throws and reports the same `IllegalArgumentException`.

All four java-tools projects compile, tests included.

Merges after the OsmAnd side.


alex-dev-neo:

PR #1689, opened 2026-09-08
Follow PrecalculatedRouteDirection and RouteConditionalHelper to OsmAnd-shared

alex-dev-neo:
Stacked on #1688, review that one first. Follows osmandapp/OsmAnd#25922.

`PrecalculatedRouteDirection` and `RouteConditionalHelper` moved from `net.osmand.router` to `net.osmand.shared.routing`. Imports only, in two files — `HHRoutingPrepareContext` also static-imports `RULE_INT_MAX` from the helper.

All four java-tools projects compile, tests included.

Merges after the OsmAnd side.


alex-dev-neo:

PR #1692, opened 2026-09-09
Follow HHRoutingConfig into net.osmand.shared.routing

alex-dev-neo:
Stacked on #1691.

`HHRoutingConfig` moved out of `HHRouteDataStructure`, where it was nested, into a class of its own in `OsmAnd-shared`, because the C++ router reads fourteen of its fields and it is one of the two arguments `nativeRouting` takes. Import changes only.

`java-tools` builds `OsmAnd-java` and `OsmAnd-shared` from source, so it has to follow. `compileJava` and `compileTestJava` pass.

Pairs with osmandapp/OsmAnd#25950 and merges after it.


alex-dev-neo:

PR #1694, opened 2026-09-10
Add MVT map objects context menu

alisa911:
## AI disclaimer

1. Study osmandapp/OsmAnd-Issues#3263, the web and the server projects and write an implementation plan; the Dev notes are recommendations, the Expected Result (tags of a vector tile object equal to search results and the mobile app) is binding. The plan: the server returns the same POI feature as search does, the web only shows it.
2. Widen the POI search radius for ways. `PoiSearchService.searchPoiByOsmId`: 11 m for a node, 600 m otherwise, because the amenity of a way or relation is stored at its centroid.
3. Remove the hardcoded transport stop tags and osm type calculation from the web, do it on the server as Android does. `TransportStopsService.isPublicTransportStop` and `findBestTransportStopId` (port of Android `TransportStopHelper`: 150/400 m, name, location and exit matching, else the nearest stop), `get-transport-stop` searches within 400 m, the POI feature gets `web_transport_stop_id`. The author asked what exactly changed on the server, why `type` became optional and whether `get-poi-by-osmid` is used only for MVT (it is not, Explore uses it with `type`).
4. A parking clicked on the map has no POI in the index (a `no_indx` type): add `get-poi-by-tags` that builds the amenity from the tile tags without duplicating code. `EntityParser.parseAmenities` and `ObfConstants.getOsmObjectId`/`getOsmEntityType(RenderedObject)` are reused after the author rejected a nested ternary and a new helper written by the agent; two constants the agent had reordered in `searchPoiByOsmId` were restored. Separate commits for the transport part and the rest.
5. Go through the review of the PR and say what is agreed. Fixed on approval: the dead `feature != null` check, the redundant `node.replaceTags`, and `get-poi-by-tags` replaced by `get-poi-by-map-object` that takes the osm id and entity type from the map object id, searches the index and falls back to the tags, so the web makes one request instead of two and `type` in `get-poi-by-osmid` is required again. Not changed with reasons: the first parsed amenity depends on the tag order as in Android `convertRenderedObjectToAmenity`, the `RenderedObject` used to decode the id is the existing API, the radius 180 vs 150 was left for the author's decision.
6. Copilot comments: the tag validation of `visible-tags` is shared with the new endpoint through `SearchController.validateTags`. The null-safety comments were declined: `MapObject.getName`/`getEnName` return an empty string, and the request body is required, so `tags` cannot be null.

The agent made no commits. The Java code was not compiled or run by the agent; the earlier version of the endpoints (`get-poi-by-osmid` without `type` and `get-poi-by-tags`) was checked by the agent against the author's deployment on test.osmand.net with the web client, the final `get-poi-by-map-object` and `validateTags` were not executed by the agent. The selenium test of the web PR was run by the author.

### DIFF
--- a/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/TransportStopsService.java
+++ b/java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/TransportStopsService.java
@@ -2,6 +2,9 @@ package net.osmand.server.api.services;
 
 import net.osmand.binary.BinaryMapIndexReader;
 import net.osmand.data.*;
+import net.osmand.osm.PoiCategory;
+import net.osmand.osm.PoiFilter;
+import net.osmand.osm.PoiType;
 import net.osmand.osm.edit.Node;
 import net.osmand.router.TransportStopsRouteReader;
 import net.osmand.server.api.services.search.MapReadersService;
@@ -26,6 +29,14 @@ public class TransportStopsService {
 	private static final int SHOW_NEARBY_ROUTES_RADIUS_METERS = 150;
 	private static final int SEARCH_STOP_RADIUS_METERS = 50;
 	private static final String KEY_NEARBY_ROUTES = "nearbyRoutes";
+	// as in Android TransportStopHelper
+	private static final int SHOW_STOPS_RADIUS_METERS = 150;
+	private static final int SHOW_SUBWAY_STOPS_FROM_ENTRANCES_RADIUS_METERS = 400;
+	private static final int MAX_DISTANCE_BETWEEN_AMENITY_AND_LOCAL_STOPS = 20;
+	private static final String TRANSPORTATION_CATEGORY = "transportation";
+	private static final String PUBLIC_TRANSPORT_FILTER = "public_transport";
+	private static final String PUBLIC_TRANSPORT_STATION_SUBTYPE = "public_transport_station";
+	private static final List<String> SUBWAY_ENTRANCE_SUBTYPES = List.of("subway_entrance", PUBLIC_TRANSPORT_STATION_SUBTYPE);
 
 	private static List<LatLon> bboxAroundPoint(double lat, double lon, int radiusMeters) {
 		QuadRect rect = MapUtils.calculateLatLonBbox(lat, lon, radiusMeters);
@@ -109,7 +120,7 @@ public class TransportStopsService {
 	}
 
 	public GeojsonClasses.Feature getTransportStop(LatLon transportStopCoords, long stopId) throws IOException {
-		List<LatLon> bbox = bboxAroundPoint(transportStopCoords.getLatitude(), transportStopCoords.getLongitude(), SEARCH_STOP_RADIUS_METERS);
+		List<LatLon> bbox = bboxAroundPoint(transportStopCoords.getLatitude(), transportStopCoords.getLongitude(), SHOW_SUBWAY_STOPS_FROM_ENTRANCES_RADIUS_METERS);
 		TransportStopsReaderResult readerResult = getTransportStopsReader(bbox);
 		if (readerResult == null) {
 			return null;
@@ -124,6 +135,67 @@ public class TransportStopsService {
 			osmAndMapsService.unlockReaders(readerResult.readers);
 		}
 		return null;
+	}
+
+	public boolean isPublicTransportStop(Amenity amenity) {
+		PoiCategory category = amenity.getType();
+		if (!TRANSPORTATION_CATEGORY.equals(category.getKeyName())) {
+			return false;
+		}
+		PoiFilter filter = category.getPoiFilterByName(PUBLIC_TRANSPORT_FILTER);
+		if (filter == null) {
+			return false;
+		}
+		for (PoiType type : filter.getPoiTypes()) {
+			if (type.getKeyName().equals(amenity.getSubType())) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	public Long findBestTransportStopId(Amenity amenity) throws IOException {
+		LatLon loc = amenity.getLocation();
+		boolean isSubwayEntrance = SUBWAY_ENTRANCE_SUBTYPES.contains(amenity.getSubType());
+		int radius = isSubwayEntrance ? SHOW_SUBWAY_STOPS_FROM_ENTRANCES_RADIUS_METERS : SHOW_STOPS_RADIUS_METERS;
+		TransportStopsReaderResult readerResult = getTransportStopsReader(bboxAroundPoint(loc.getLatitude(), loc.getLongitude(), radius));
+		if (readerResult == null) {
+			return null;
+		}
+		try {
+			List<TransportStop> stops = new ArrayList<>(readerResult.transportReaders.readMergedTransportStops(readerResult.request));
+			stops.sort(Comparator.comparingDouble(s -> MapUtils.getDistance(s.getLocation(), loc)));
+			for (TransportStop stop : stops) {
+				if (isSubwayEntrance ? isStopOfStation(stop, amenity) : isStopOfAmenity(stop, amenity)) {
+					return stop.getId();
+				}
+			}
+			return stops.isEmpty() ? null : stops.get(0).getId();
+		} finally {
+			osmAndMapsService.unlockReaders(readerResult.readers);
+		}
+	}
+
+	private static boolean isStopOfAmenity(TransportStop stop, Amenity amenity) {
+		LatLon loc = amenity.getLocation();
+		String stopName = stop.getName().toLowerCase();
+		String amenityName = amenity.getName().toLowerCase();
+		boolean sameName = stopName.contains(amenityName) || amenityName.contains(stopName);
+		return (sameName && MapUtils.getDistance(stop.getLocation(), loc) < MAX_DISTANCE_BETWEEN_AMENITY_AND_LOCAL_STOPS)
+				|| stop.getLocation().equals(loc);
+	}
+
+	private static boolean isStopOfStation(TransportStop stop, Amenity amenity) {
+		if (PUBLIC_TRANSPORT_STATION_SUBTYPE.equals(amenity.getSubType())
+				&& (stop.getName().equals(amenity.getName()) || stop.getEnName(false).equals(amenity.getEnName(false)))) {
+			return true;
+		}
+		for (TransportStopExit exit : stop.getExits()) {
+			if (MapUtils.getDistance(exit.getLocation(), amenity.getLocation()) < MapUtils.ROUNDING_ERROR) {
+				return true;
+			}
+		}
+		return false;
 	}
 
 	public Map<String, Object> getNearbyTransportStops(LatLon stopCoords, long excludeStopId) throws IOException {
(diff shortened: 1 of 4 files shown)


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3263

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
```diff
@@ -492,11 +503,15 @@ private boolean equalsIgnoreCaseSafe(String s, String t) {
...
-		final double OTHER_POI_SEARCH_RADIUS = 0.0001; // ~11 meters
-		double radiusDegree = type.equals(RELATION_TYPE) ? RELATION_SEARCH_RADIUS : OTHER_POI_SEARCH_RADIUS;
-		return searchSinglePoi(loc, radiusDegree, new ResultMatcher<>() {
+		final String NODE_TYPE = "1";
+		return searchPoiByOsmId(loc, osmid, NODE_TYPE.equals(type) ? EntityType.NODE : EntityType.WAY, timeZone);
+	}
+
+	private Feature searchPoiByOsmId(LatLon loc, long osmid, EntityType type, String timeZone) throws IOException {
+		final double WAY_SEARCH_RADIUS = 0.0055; // ~600 meters, way/relation amenity is located at its centroid
+		final double NODE_SEARCH_RADIUS = 0.0001; // ~11 meters
```
RZR-UA, line 512: Please find similar constants in Android and link them here instead of reinventing local ones.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
```diff
@@ -492,11 +503,15 @@ private boolean equalsIgnoreCaseSafe(String s, String t) {
...
-		final double RELATION_SEARCH_RADIUS = 0.0055; // ~600 meters
-		final double OTHER_POI_SEARCH_RADIUS = 0.0001; // ~11 meters
-		double radiusDegree = type.equals(RELATION_TYPE) ? RELATION_SEARCH_RADIUS : OTHER_POI_SEARCH_RADIUS;
-		return searchSinglePoi(loc, radiusDegree, new ResultMatcher<>() {
+		final String NODE_TYPE = "1";
+		return searchPoiByOsmId(loc, osmid, NODE_TYPE.equals(type) ? EntityType.NODE : EntityType.WAY, timeZone);
+	}
+
+	private Feature searchPoiByOsmId(LatLon loc, long osmid, EntityType type, String timeZone) throws IOException {
+		final double WAY_SEARCH_RADIUS = 0.0055; // ~600 meters, way/relation amenity is located at its centroid
```
RZR-UA, line 511: Check on immense objects, such as the Mediterranean Sea (35.000754, 20.000610). Neither Android nor Web displays tags for seas and the like.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
```diff
@@ -506,12 +521,47 @@ public boolean publish(Amenity amenity) {
...
 			}
-		}, timeZone);
+		});
+		return res != null ? getMapObjectFeature(res, timeZone) : null;
+	}
+
+	private Feature getMapObjectFeature(SearchResult res, String timeZone) throws IOException {
+		Feature feature = searchResultConverter.getPoiFeature(res, timeZone);
+		Amenity amenity = (Amenity) res.object;
+		if (transportStopsService.isPublicTransportStop(amenity)) {
```
RZR-UA, line 531: Should be simplified to if (res.object instanceof Amenity amenity && ...)

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
```diff
@@ -506,12 +521,47 @@ public boolean publish(Amenity amenity) {
...
+	// vector tile object: POI from the index by its osm id, else (no_indx types, buildings, ...) an amenity from its tags as the map creator does
+	public Feature getPoiByMapObject(long mapObjectId, LatLon loc, Map<String, String> tags, String timeZone) throws IOException {
+		RenderedObject renderedObject = new RenderedObject();
+		renderedObject.setId(mapObjectId);
+		long osmId = ObfConstants.getOsmObjectId(renderedObject);
+		EntityType entityType = ObfConstants.getOsmEntityType(renderedObject);
+		Feature feature = searchPoiByOsmId(loc, osmId, entityType, timeZone);
+		if (feature != null || tags.isEmpty()) {
+			return feature;
+		}
```
RZR-UA, line 549: Potential NPE for caller if tags.isEmpty(). Please refactor this block.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
```diff
@@ -506,12 +521,47 @@ public boolean publish(Amenity amenity) {
...
+		MapPoiTypes poiTypes = poiTypesService.getMapPoiTypes(PoiTypesService.DEFAULT_SEARCH_LANG);
+		Node node = new Node(loc.getLatitude(), loc.getLongitude(), -1);
+		List<Amenity> amenities = EntityParser.parseAmenities(poiTypes, node, tags, new ArrayList<>(), false);
+		if (amenities.isEmpty()) {
+			return null;
+		}
+		Amenity amenity = amenities.get(0);
```
RZR-UA, line 556: Shall we log amenities.size() > 1 here? At least add a comment about a possible future implementation for the What's here menu on the web.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
```diff
@@ -506,12 +521,47 @@ public boolean publish(Amenity amenity) {
...
+		EntityType entityType = ObfConstants.getOsmEntityType(renderedObject);
+		Feature feature = searchPoiByOsmId(loc, osmId, entityType, timeZone);
+		if (feature != null || tags.isEmpty()) {
+			return feature;
+		}
+		MapPoiTypes poiTypes = poiTypesService.getMapPoiTypes(PoiTypesService.DEFAULT_SEARCH_LANG);
+		Node node = new Node(loc.getLatitude(), loc.getLongitude(), -1);
+		List<Amenity> amenities = EntityParser.parseAmenities(poiTypes, node, tags, new ArrayList<>(), false);
+		if (amenities.isEmpty()) {
+			return null;
```
RZR-UA, line 554: How does the caller check for null from this method? Consider @Nullable @NonNull, etc

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
```diff
@@ -506,12 +521,47 @@ public boolean publish(Amenity amenity) {
...
+		if (transportStopsService.isPublicTransportStop(amenity)) {
+			Long stopId = transportStopsService.findBestTransportStopId(amenity);
+			if (stopId != null) {
+				feature.prop(SearchResultConverter.PoiTypeField.TRANSPORT_STOP_ID.getFieldName(), stopId);
+			}
+		}
+		return feature;
+	}
+
+	// vector tile object: POI from the index by its osm id, else (no_indx types, buildings, ...) an amenity from its tags as the map creator does
```
RZR-UA, line 540: Please check the nameless object at `50.419108, 30.504983`. The web opens an empty menu, whereas Android displays a type=tower icon.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/TransportStopsService.java
```diff
@@ -126,6 +137,67 @@ public GeojsonClasses.Feature getTransportStop(LatLon transportStopCoords, long
...
+			for (TransportStop stop : stops) {
+				if (isSubwayEntrance ? isStopOfStation(stop, amenity) : isStopOfAmenity(stop, amenity)) {
+					return stop.getId();
+				}
+			}
+			return stops.isEmpty() ? null : stops.get(0).getId();
+		} finally {
+			osmAndMapsService.unlockReaders(readerResult.readers);
+		}
+	}
```
RZR-UA, line 177: Does this approach match any of Android's? Why not all? What is **"best"** for a random user? Why does SUBWAY_ENTRANCE_SUBTYPES affect the radius? Consider removing the TransportStop methods from this PR, since the issue didn't request them.

At any rate, if you copy Android's findBestTransportStopForAmenity() approach, consider sharing the method between Android and Web.

The final decision is on @vshcherb, but IMHO it is a bad idea to lock/use readers to display the Web context menu for map clicks.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/TransportStopsService.java
```diff
@@ -26,6 +29,14 @@ public class TransportStopsService {
 	private static final int SHOW_NEARBY_ROUTES_RADIUS_METERS = 150;
 	private static final int SEARCH_STOP_RADIUS_METERS = 50;
 	private static final String KEY_NEARBY_ROUTES = "nearbyRoutes";
+	// as in Android TransportStopHelper
+	private static final int SHOW_STOPS_RADIUS_METERS = 150;
+	private static final int SHOW_SUBWAY_STOPS_FROM_ENTRANCES_RADIUS_METERS = 400;
+	private static final int MAX_DISTANCE_BETWEEN_AMENITY_AND_LOCAL_STOPS = 20;
```
RZR-UA, line 35: Please refactor Web/Android to use shared consts and please them in OsmAnd-java. It's a bad approach to duplicate consts and even name them differently (e.g., SHOW_STOPS_RADIUS_METERS vs SHOW_STOPS_RADIUS_METERS_UI)

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
```diff
@@ -351,6 +361,21 @@ public ResponseEntity<String> getPoiByOsmId(@RequestParam double lat,
...
+	                                                @RequestParam double lat,
+	                                                @RequestParam double lon,
+	                                                @RequestParam long id,
+	                                                @RequestParam(required = false) String timeZone) throws IOException {
+		ResponseEntity<String> error = validateTags(tags);
+		if (error != null) {
+			return error;
+		}
+		Feature poi = poiSearchService.getPoiByMapObject(id, new LatLon(lat, lon), tags, timeZone);
+		return ResponseEntity.ok(gson.toJson(poi));
```
RZR-UA, line 376: getPoiByMapObject() is @Nullable, NPE risk

RZR-UA (changes requested):

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/SearchController.java
alisa911: No NPE, Gson toJson returns null for a null argument and the web already treats a null feature as not found.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
alisa911: tower

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
alisa911: sea

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/search/PoiSearchService.java
alisa911: nav-final-2

RZR-UA (approved):

PR #1698, opened 2026-09-11
Add FastSpring tests

alisa911:
## AI disclaimer

1. Explain the Jenkins `UpdateSubscription` log lines for a FastSpring purchase. Mapped each line to `UpdateSubscription.queryPurchases` / `updateSubscriptionDb`; the FastSpring rows were only skipped there as too recent (15 minutes), nothing was written by the job.
2. Find why the FastSpring subscription became `valid=false` with an empty `kind` although the hook stores `valid=true`. Traced the only writer that leaves `kind` empty to `UserSubscriptionService.revalidateFastSpringSubscription`, which runs on every user request while `checktime` is null; the server log confirmed `has null nextChargeDate` 16 minutes after purchase. The user had canceled the subscription in the FastSpring portal, and the real `GET /subscriptions/{id}` response for a canceled subscription has `active=true`, `state=canceled`, `autoRenew=true` and no `nextChargeDate`, only `next`, `end`, `deactivationDate`. The FastSpring docs do not describe this; the author verified it against the live API.
3. Audit how FastSpring subscriptions are handled on the server and whether unit tests are possible. Published an artifact covering hooks, revalidation, the job, configuration, findings and testability; the internal doc `OsmAnd-Issues/Development/FastSpring.md` was found in branch `250826_1` and checked against the code.
4. Write tests first with the fix reverted, using real API responses as separate JSON files. Added `OsmAndServer/src/test/java/net/osmand/server/fs` (JUnit 4 + Mockito, no Spring context) and anonymized fixtures in `src/test/resources/fs` for `GET /subscriptions` active/canceled/deactivated, `GET /orders` completed/refunded, `order.completed` and `return.created` for a subscription and an in-app. The canceled test failed on the old code as expected; the author ran the tests. Removed the unused `wiremock-standalone` test dependency (2018) whose bundled slf4j 1.7 shadowed slf4j 2 in the test classpath.
5. Apply the fix. `FastSpringSubscription` reads `next`, `deactivationDate`, `state`; `getExpiryTime()` returns `deactivationDate` for canceled/deactivated and `next` otherwise; `isAutoRenewing()` follows `state` because FastSpring keeps `autoRenew=true` after cancel; `toSubscriptionPurchase()` extracted from `UpdateSubscription.processFastSpringSubscription`. `revalidateFastSpringSubscription` keeps a canceled subscription valid until `expiretime` taken from FastSpring. `FastSpringOrder` reads `returns[]`, so a refunded order (which stays `completed=true`) is invalid for in-app validation in `UpdateInAppPurchase`.
6. Check the refund handling against the FastSpring docs and add tests for hook error branches. `revokeSubscriptionsIfDeactivated` matches the documented "Cancel Related Subscriptions" behavior; added tests for duplicate order, unknown user (200, nothing to retry since checkout requires login), unknown SKU (202) and refund of an unrecorded order (202).
7. Review findings: inactive branch left `autorenewing=true` (fixed), `isCanceled()` renamed to `isCanceledOrDeactivated()`, refund of a canceled subscription now also syncs `expiretime`, `json()` moved from a test class into `FsJson`.

No commits were made by the agent; the tests were run by the author. The module was not built in the agent's environment (no access to Maven Central), only `FastSpringHelper` was compiled in isolation. Webhook signature verification and re-checking of `valid=false` rows were noted in the audit but not changed.


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3311

alisa911:
## Fixed

1. Canceled subscription was invalidated immediately: FastSpring returns `active=true, state=canceled` without `nextChargeDate`; revalidation treated it as dead. Now valid until `deactivationDate`.
2. `autorenewing` stayed `true` after cancel (FastSpring keeps `autoRenew=true`); now derived from `state`.
3. Inactive branch left `autorenewing=true` on a `valid=false` record.
4. Refund of a canceled subscription did not sync `expiretime` from FastSpring.
5. Refunded in-app was valid on job revalidation: order stays `completed=true`, refund is only in `returns[]`, which was not read.
6. `FastSpringPurchase.isValid()` NPE on `completed=null`.
7. Unused `wiremock-standalone` shadowed slf4j 2 in the test classpath.

## Recommendation

Subscribe to `subscription.canceled`, `subscription.deactivated` and `subscription.charge.completed`. Today cancel, deactivation and renewal reach the DB only via revalidation (job: 14 days for an active subscription; server: on the first user request after `expiretime` passes), so `autorenewing`, `expiretime` and `price` lag behind FastSpring for up to two weeks.

alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3334

RZR-UA (approved):

PR #1708, opened 2026-09-13
OsmGpx classifier: piste is not a skiing keyword

vshcherb:
`piste` (a skiing tag in activities.json) is an ordinary word in French: *piste cyclable* is a cycle path, *piste agricole* a farm track, *piste des douaniers* a coastal path. Found as a keyword in tags, names and descriptions, it labelled bike rides as skiing, and the wide winter_sport speed envelope (p85 up to 120 km/h) accepted them. Example: [trace 3322688](https://www.openstreetmap.org/user/LeFanch/traces/3322688), "Boucle Biganos - Andernos via Lubec, retour par piste cyclable", 22 km/h in July at sea level, stored as skiing.

`piste` and `pistes` join `STOP_KEYWORDS`. `piste:type` and `piste:difficulty` stay keywords.

Checked on the 30K validation sample by re-running the keyword and speed steps for every track with `piste` in its tags, name or description:
- The simulation reproduces all 25 stored labels.
- Skiing from `piste` went from 21 to 5 tracks. The 5 real ski tracks (Les Houches, San Bernardino, Hlíðarfjall ×2, Pallas) stay skiing through `ski`, `skiing` or `downhill`.
- The other 16 move to cycling, mountain_biking or road_cycling (11), foot (2), driving (2), or Unknown activity (1, no timing), from the description or from speed.

On the full database (stats export of 2026-09-13, 6,525 skiing tracks):
- 1,067 skiing tracks mention `piste`.
- 721 of them have no ski word (ski, snowboard, langlauf, nordic, lift, …); this change is expected to move about these 721.
- 679 are named `Piste_<date>.gpx`, the default track name of French-language GPS firmware; 410 of those were recorded May–October in the northern hemisphere.

The change takes effect on the next classify run.

## AI disclaimer

Prompts used (summarised):
1. Trace 3322688 is stored as skiing — why, and is it really skiing?
2. Isn't "piste" skiing?

Decided by the agent, not requested explicitly: stopping `piste`/`pistes` rather than removing them from activities.json (the OSM `piste:*` keys stay); the 30K sample check.


vshcherb:
Moved into #1706 together with the high-speed train fix, so the classifier changes go in one PR. (gen by AI)

PR #1712, opened 2026-09-14
Handle FastSpring subscription webhooks: canceled, deactivated, charge.completed

alisa911:
Adds three FastSpring subscription webhooks. Until now a cancellation, a deactivation or a renewal reached `supporters_device_sub` only through revalidation: the Jenkins job touches an active subscription once per 14 days, the server only after `expiretime` has passed. So `autorenewing`, `expiretime` and `valid` lagged behind FastSpring by up to two weeks.

Refs #3334.

### Endpoints

| Event | Endpoint | Fires |
|---|---|---|
| `subscription.canceled` | `/fs/subscription-canceled` | customer portal cancel, "Deactivate at Next Period", API cancel without `billingPeriod=0` |
| `subscription.deactivated` | `/fs/subscription-deactivated` | end of the paid period, "Deactivate Now", API cancel with `billingPeriod=0` |
| `subscription.charge.completed` | `/fs/subscription-charge-completed` | automatic and managed rebills, proration adjustments (not the initial purchase) |

All three have to be configured in the FastSpring dashboard (event -> URL), otherwise they are never delivered.

### How the record is found

The `subscription.*` payloads carry the subscription object, not the order: there is no order id in them, while our record is keyed by `orderid + sku`. `subscription.charge.completed` does carry an order, but it is the rebill order, not the one we store. So all three handlers take only the subscription id from the payload, read `GET /subscriptions/{id}` and use `initialOrderId` + `sku` from the response. A rebill therefore updates the existing record instead of inserting a second one under the new order id.

Missing record (the hook overtook `order.completed`) is answered with an error so that FastSpring retries, same as `return.created`.

### State

The state is applied by `UserSubscriptionService.applyFastSpringSubscription`, extracted from `revalidateFastSpringSubscription`, so the hook and the revalidation cannot diverge:

- canceled: `autorenewing=false`, `expiretime=deactivationDate`, stays valid until then
- deactivated: `valid=false`, `autorenewing=false`, `kind=expired` unless refund or chargeback already set it (previously the server revoked a subscription with an empty `kind`)
- rebill: `expiretime` moves to the next billing date

The hook also sets `checktime`, unlike revalidation: the data comes straight from the API, so neither the job nor the next user request has to re-check it.

### Tests

`FastSpringSubscriptionEventsTest`, 4 cases. The webhook fixtures are built from the documented payloads — no real events have been recorded yet; the API responses they are validated against are real. `FastSpringSubscriptionsGetTest` gains an assertion on `kind`.

## AI disclaimer

Written with Claude Code (Claude Opus 5), driven by [@alisa911](https://github.com/alisa911). Summary of the requests it worked from, in order:

1. Continue the FastSpring task from a handoff file and issue #3334: add the three remaining subscription webhooks. The agent read the merged work from #1698, proposed a plan, and was told to stop planning and take the hooks one at a time, reading the FastSpring documentation for each.

2. Add subscription.canceled. The documented payload turned out to carry the subscription object only, with no order id, while our record is keyed by orderid plus sku, so the handler takes the subscription id from the payload and reads GET /subscriptions/{id} for initialOrderId and the current state. FastSpringHelper.getSubscription became public and gained initialOrderId, and the state mapping was extracted from revalidateFastSpringSubscription into applyFastSpringSubscription so the hook and the revalidation cannot diverge.

3. Corrections on process and naming: never commit without explicit permission, keep only real recorded payloads in fixtures (the doc-based test was removed, then restored on the author's decision with a note that no real event exists yet), and name the endpoint after the event, so /fs/subscription became /fs/subscription-canceled following /fs/order-completed.

4. Explain why the build failed. The agent found that java-tools/settings.gradle builds against the neighbouring ~/osmand/android checkout, which was 42 commits behind and missing SpatialSearchResult.getAtoms(); the author updated it and compilation passed.

5. Add subscription.deactivated. While checking who writes the kind column, the agent reported that the server revoked subscriptions leaving kind empty; the author said a found bug is to be fixed, not documented, so deactivation now writes UpdateSubscription.EXPIRED_STATE (made public instead of declaring a new constant) unless refund or chargeback already set kind.

6. Add subscription.charge.completed. A rebill is a new order, so the handler resolves initialOrderId and updates the existing record instead of inserting a second one under the rebill order id; the documentation for order.completed confirms it does not fire for rebills.

7. Review the code-review findings. Confirmed by the agent and fixed: a record revoked with kind refund or chargeback was re-validated back to valid, deactivation left expiretime at the old billing date while getSubscriptionInfo derives the user-visible state from it, and an event without a subscription id was acknowledged instead of retried. The finding about events for subscriptions that are absent from the database was dismissed by the author as impossible in practice.

8. Repeated pushback on the agent's own changes: the kind constants were moved around twice before landing in FastSpringHelper next to SUBSCRIPTION_STATE_*, the collapsed dispatch branches were restored to three explicit ones, and a duplicated documentation link was removed.

The agent made no commits without explicit permission and pushed only when asked. The FastSpring tests (net.osmand.server.fs, 21 cases) were run locally by the agent with gradle; the android checkout update needed to compile the project was done by the author. The webhook fixtures are built from the documented payloads, no live events have been recorded, and the three endpoints still have to be enabled in the FastSpring dashboard, which the agent has no access to.


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3334

RZR-UA (approved):

PR #1717, opened 2026-09-15
New email templates for Web

Dima-1:
https://github.com/osmandapp/OsmAnd-Issues/issues/2676
## AI disclaimer

Written with Claude Code (Claude Opus 5), driven by [@Dima-1](https://github.com/Dima-1). Summary of the requests it worked from, in order:

1. Explain how notification emails are sent and which templates exist in web-server-config, then suggest what to do for OsmAnd-Issues#2676 and turn Phase 1 into a plan with exact files. The agent traced `EmailSenderTemplate` (comment directives, `defaults.html`, unsubscribe include) and proposed a shared base template for the 10 transactional emails, with the subject and H1 of each email matching the issue checklist.
2. Implement it, then questions on the design: why `USE_BASE`/`TRANSACTIONAL` are not defaults, why `EMAIL_TITLE` duplicates the H1, why styles repeat in every template, and whether `INTRO_NEW` is used. Result: legacy templates keep `defaults.html` untouched; new ones opt in via `USE_BASE` and get `base.html`, `base-locale/<lang>.html`, `header.html` and `footer.html`; shared styles moved into `base.html`.
3. "Is there no flow for password changing?" and several "Review uncommitted changes" rounds with selected fixes. Added `CloudAccountAction` (setup, password, change-request, change, delete, email-changed); the old address now gets a warning with a masked new address (`maskEmail`) when the change is requested; the purchase receipt (`sendPurchaseReceiptEmail`) shows Renews on, Expires on or no row depending on the SKU; share approved/declined emails are sent from `ShareFileService.notifyAccessRequest` without exposing the owner's email.
4. How to make buttons render in Outlook (VML), pushback "why so complicated", and "make it without restoreMsoMarkers", then "isProtectedMsoSpan vs isMsoMarkerLine, is one enough?". Replaced the protect/restore round-trip with a single pass `stripCommentsKeepingMso` + `isProtectedMsoSpan`, which also stops `<!--[if !mso]><!-->` becoming a bogus `if:` SMTP header. Covered by `MsoConditionalCommentTest`.
5. "Only command comments - may this be a bug?". The agent found that any unrecognized `<!--X: Y-->` comment becomes an SMTP header (9 junk headers from `blackfriday/index.html`). An allowlist plan was proposed and not approved, so it is not in this PR.
6. Check consistency with the issue and list the open questions, add translations. Added `cloud/account/*` in 11 languages with `base-locale` strings (rtl for ar), and `EmailRenderTest` rendering every template and locale into `build/email-preview/`.
7. Careful and skill-based reviews of the full diff, with selected fixes: `safeLang` against path traversal through `lang`, HTML/`@` escaping of user values (`htmlText`, `plainText`), the receipt sent only after commit to avoid duplicates on FastSpring retry, escaped `NEW_EMAIL`, 10-minute code expiry. The agent also corrected its own wrong claims (CI failing, header count, "removing access sends Request declined"). Two later fixes (wear-out fraction, ownership check in `updateRequests`) were undone at the author's request.
8. Remove non-meaningful comments from the diff, and change the logo URL to `https://osmand.net/img/logo_osmand_black-text.png`. The agent reported that this URL currently returns 404.

Code was compiled and the email render tests were run locally by the agent (the build has `ignoreFailures = true`, so results were checked in the test reports). Rendered emails were not checked in real mail clients, and no commits were made by the agent.


java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -1517,7 +1538,7 @@ public ResponseEntity<String> confirmCode(String username, String token) {
         if (pu == null) {
             return ResponseEntity.badRequest().body("User is not registered");
         }
-        boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() > TimeUnit.MILLISECONDS.convert(24, TimeUnit.HOURS);
+        boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() > CODE_EXPIRATION_TIME_MS;
         wearOutToken(pu);
```
alisa911: `wearOutToken` still subtracts 7 hours, which was tuned for the 24h TTL. With `CODE_EXPIRATION_TIME_MS` = 10 min a single wrong attempt expires the code, so the user has to request a new one after one typo.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -607,12 +611,12 @@ public ResponseEntity<String> registerNewDevice(String email, String token, Stri
...
-        if ( ! (tokenIsActive && pu.token.equals(token))) {
+        long tokenAge = pu.tokenTime == null ? Long.MAX_VALUE : System.currentTimeMillis() - pu.tokenTime.getTime();
+        boolean codeIsActive = pu.token != null && pu.tokenTime != null && tokenAge < CODE_EXPIRATION_TIME_MS;
+        boolean passwordIsAllowed = pu.token != null && pu.tokenTime != null && tokenAge < WEB_PASSWORD_WINDOW_MS;
+        if ( ! (codeIsActive && pu.token.equals(token))) {
             wearOutToken(pu); // cut down on tries (even for web password)
-            if ( ! (tokenIsActive && validateWithWebPassword(pu.id, token))) {
+            if ( ! (passwordIsAllowed && validateWithWebPassword(pu.id, token))) {
                 LOG.error("device-register: invalid token (" + email + ") [" + token + "]");
                 throw new OsmAndPublicApiException(ERROR_CODE_TOKEN_IS_NOT_VALID_OR_EXPIRED, "token is not valid or expired (24h)");
```
alisa911: Error text still says "expired (24h)" while the TTL is now 10 minutes.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -1420,7 +1424,7 @@ public void getBackupFolder(HttpServletResponse response, CloudUserDevicesReposi
     public ResponseEntity<String> deleteAccount(String token, CloudUserDevicesRepository.CloudUserDevice dev, HttpServletRequest request) throws ServletException {
         CloudUsersRepository.CloudUser pu = usersRepository.findById(dev.userid);
         if (pu != null && pu.id == dev.userid) {
-            boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() > TimeUnit.MILLISECONDS.convert(24, TimeUnit.HOURS);
+            boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() > CODE_EXPIRATION_TIME_MS;
             boolean validToken = pu.token.equals(token) && !tokenExpired;
             wearOutToken(pu);
```
alisa911: Line 1448: error text still says "expired (24h)" while the TTL is now 10 minutes.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -1500,7 +1521,7 @@ public ResponseEntity<String> confirmCode(String code, CloudUserDevicesRepositor
         if (pu == null) {
             return ResponseEntity.badRequest().body("User is not registered");
         }
-        boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() > TimeUnit.MILLISECONDS.convert(24, TimeUnit.HOURS);
+        boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() > CODE_EXPIRATION_TIME_MS;
         wearOutToken(pu);
         if (pu.token.equals(code) && !tokenExpired) {
             return ok();
```
alisa911: Line 1529: error text still says "expired (24h)" while the TTL is now 10 minutes.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -13,8 +13,11 @@ public static void test(String[] args) {
 		EmailSenderService sender = new EmailSenderService();
 		String email = "dmarc-reports@osmand.net";
 		sender.sendOsmAndCloudPromoEmail(email, "promo");
-		sender.sendOsmAndCloudWebEmail(email, "token", "en", "delete");
+		sender.sendOsmAndCloudWebEmail(email, "token", "delete", "en");
```
alisa911: `sendOsmAndCloudWebEmail` and the `cloud/web/*` templates are only used from this `main()` now; remove them or note why they stay.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/FastSpringController.java
```diff
@@ -431,6 +446,58 @@ private void setInitialSubscriptionDates(DeviceSubscriptionsRepository.Supporter
...
+	                                 List<DeviceInAppPurchasesRepository.SupporterDeviceInAppPurchase> purchases,
+	                                 List<DeviceSubscriptionsRepository.SupporterDeviceSubscription> subscriptions) {
+		FastSpringWebhookRequest.Data data = event.data;
+		try {
+			String productName;
+			String planName;
+			String renewalLabel;
+			Date renewalDate;
+			if (!subscriptions.isEmpty()) {
+				DeviceSubscriptionsRepository.SupporterDeviceSubscription sub = subscriptions.get(0);
```
alisa911: Only the first item of the order goes into the receipt; an order with several items loses the rest.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/ShareFileService.java
```diff
@@ -258,11 +261,53 @@ public boolean updateRequests(Map<Integer, String> accessMap) {
 			if (access != null) {
 				access.access = (accessType);
 				shareFileRepository.saveAndFlush(access);
+				if (PermissionType.READ.name().equals(accessType)) {
+					notifyAccessRequest(access, true);
+				} else if (PermissionType.BLOCKED.name().equals(accessType)) {
+					notifyAccessRequest(access, false);
```
alisa911: Emails are sent inside the transaction; on rollback the requester has already been told the access was approved. The afterCommit approach used in `FastSpringController` would fit here too.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
alisa911: error text still says "expired (24h)" while the TTL is now 10 minutes.

alisa911, line 1556: wearOutToken still subtracts 7 hours, which was tuned for the 24h TTL. With CODE_EXPIRATION_TIME_MS = 10 min a single wrong attempt now expires the code, so the user has to request a new one after one typo.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
alisa911, line 621: Error text still says "expired (24h)" while the TTL is now 10 minutes.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
alisa911, line 1448: Same here: "(24h)" is no longer true.

alisa911, line 1529: Same here: "(24h)" is no longer true.

alisa911, line 1546: Same here: "(24h)" is no longer true.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -132,6 +132,8 @@ public class UserdataService {
     public static final String MODEL_DEVICE_WEB = "Web";
     public static final String TOKEN_DEVICE_WEB = "web";
     public static final int ERROR_CODE_PRO_USERS = 100;
+    public static final long CODE_EXPIRATION_TIME_MS = TimeUnit.MILLISECONDS.convert(10, TimeUnit.MINUTES);
```
alisa911, line 135: This also shortens the code lifetime for the Android/iOS device registration flow, not only for the web emails. Is that intended?

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +54,101 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+	}
+
+	// The same for a plain-text header such as Subject, where entities would show literally: no line breaks,
+	// and '@' replaced with the full-width look-alike so it cannot form an @VAR@ token.
+	static String plainText(String s) {
+		return s == null ? "" : s.replaceAll("[\\r\\n]+", " ").replace('@', '＠');
+	}
+
+	// cloud/purchase/receipt. renewalLabel/renewalDate == null means a lifetime purchase: the Renews/Expires row is hidden.
+	public void sendPurchaseReceiptEmail(String email, String orderId, String orderDate, String orderTotal,
```
alisa911, line 120: There is no sender for cloud/purchase/linked (email 6 in the issue); the template in web-server-config#210 is never used.

alisa911, line 120: EMAIL, ORDER_ID, ORDER_TOTAL and PRODUCT_NAME come from the FastSpring payload and go into HTML unescaped, unlike the share-file values. Either escape them too or move escaping into EmailSenderTemplate.set().

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +54,101 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+	// User-supplied text placed into an HTML template: escape markup, and '@' as &#64;, so the value can neither
+	// inject HTML into an OsmAnd-branded email nor be expanded (or rejected) by the template's @VAR@ substitution.
+	static String htmlText(String s) {
+		return s == null ? "" : org.springframework.web.util.HtmlUtils.htmlEscape(s).replace("@", "&#64;");
+	}
+
+	// The same for a plain-text header such as Subject, where entities would show literally: no line breaks,
+	// and '@' replaced with the full-width look-alike so it cannot form an @VAR@ token.
+	static String plainText(String s) {
+		return s == null ? "" : s.replaceAll("[\\r\\n]+", " ").replace('@', '＠');
```
alisa911, line 116: This changes the file name in the subject (a@b.gpx becomes a＠b.gpx). Subject is not run through @VAR@ substitution after set(), so a plain @ should be fine here; if not, escape only the @[A-Z_]+@ pattern.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/ShareFileService.java
```diff
@@ -258,11 +261,53 @@ public boolean updateRequests(Map<Integer, String> accessMap) {
...
+
+	private void notifyAccessRequest(ShareFileRepository.ShareFilesAccess access, boolean approved) {
+		try {
+			ShareFileRepository.ShareFile file = access.file;
+			CloudUsersRepository.CloudUser requester = access.user;
+			if (file == null || requester == null || requester.email == null) {
+				return;
+			}
+			CloudUsersRepository.CloudUser owner = usersRepository.findById(file.ownerid);
+			String ownerName = owner != null && owner.nickname != null ? owner.nickname : "The file owner";
```
alisa911, line 284: With no nickname the email reads "The file owner approved your request to access this file". Should this fall back to the owner's email instead?

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
```diff
@@ -414,10 +476,36 @@ private void parse(List<String> lines) {
...
-		String joined = String.join("", bodyLines).replaceAll(HTML_COMMENT_REPLACE, "");
+		String joined = stripCommentsKeepingMso(String.join("", bodyLines));
 		body = body == null ? joined : body + joined; // concat bodies from all included templates
 	}
 
+	private String stripCommentsKeepingMso(String text) {
+		if (!"true".equals(vars.get(USE_BASE))) {
+			return text.replaceAll(HTML_COMMENT_REPLACE, "");
+		}
+		Matcher m = Pattern.compile(HTML_COMMENT_REPLACE).matcher(text);
```
alisa911, line 487: Compile this pattern once (static final); it runs per template line.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
```diff
@@ -47,6 +47,29 @@ Check findTemplateFile() to see how files are loaded from EMAIL_TEMPLATES direct
...
+	gets mangled by the generic comment handling below.
+
+	@TRANSACTIONAL@ - user-defined option to mark an email as transactional (verification codes,
+		account security, file-sharing notifications). Such emails cannot be opted out of, so
+		unsubscribe.html is not included for them. Set it in the template itself:
+			<!--Set TRANSACTIONAL=true-->
+
 Template variables:
 
 	The template engine supports 1st and 2nd level variables (e.g. FIRST=1 SECOND=@FIRST@ THIRD=@SECOND@)
```
alisa911, line 75: Stale: fill() now does 4 passes, so up to 4 levels are supported.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/FastSpringController.java
```diff
@@ -431,6 +446,58 @@ private void setInitialSubscriptionDates(DeviceSubscriptionsRepository.Supporter
 		}
 	}
 
+	private void sendPurchaseReceipt(String email, FastSpringWebhookRequest.Event event,
```
alisa911, line 449: Where does the data for the receipt come from? totalDisplay is a new field on the webhook payload that nothing else in the project uses, productName comes from PurchasesDataLoader, and the plan/renewal labels are built here. Please point to the FastSpring docs for totalDisplay and confirm it is present in order.completed.

Also, all receipt strings are hardcoded English in the controller ("Annual subscription", "One-time purchase", "Renews on", "Expires on", date format MMM d, yyyy). Account emails are localized via templates; why is the receipt different, and why is this presentation logic in the controller instead of the template or the sender?

alisa911 (commented):
The README in web-server-config#210 refers to MsoConditionalCommentTest, but this PR adds no tests.

The PR reaches into FastSpringController and ShareFileService, which is outside the email templates scope. Sending logic and the new helpers (readableFileSize, sendPurchaseReceipt, plan/label building) belong in EmailSenderService or a dedicated service, not in the controller and the share service.

CODE_EXPIRATION_TIME_MS and WEB_PASSWORD_WINDOW_MS are dropped into UserdataService without explanation, and the TTL change also affects the Android/iOS device registration flow. Please explain why they live there and why 24h → 10 min for all flows.

We have EmailSenderService, yet sendCode and the action → template mapping live in UserdataService. Why is this split?

Actions are still string literals ("delete", "change", "setup") compared inline; there is a CloudAccountAction enum now, use it (or constants) end to end.
readableFileSize duplicates the private size() in AnalyzeRunOperation; either share one util or reuse the existing one.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/c0f121aba7baecc5e43266810ac14ad07e31bcff

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/c0f121aba7baecc5e43266810ac14ad07e31bcff

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/c0f121aba7baecc5e43266810ac14ad07e31bcff

Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/c0f121aba7baecc5e43266810ac14ad07e31bcff

Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/c0f121aba7baecc5e43266810ac14ad07e31bcff

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
Dima-1: Yes, intended and agreed with the product owner: the code lifetime is 10 minutes for every flow.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/720766353ef7b89c7092e8b38ac49f9a4320eb9c

Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/af8dae3bfbe840d00169030e1d3a4c90a61a1a9b

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/af8dae3bfbe840d00169030e1d3a4c90a61a1a9b

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/af8dae3bfbe840d00169030e1d3a4c90a61a1a9b

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/af8dae3bfbe840d00169030e1d3a4c90a61a1a9b

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/FastSpringController.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/54a9ba4483c8e2e9212c858a48ebec7a7decfb75

Dima-1: totalDisplay is a documented order-level field of order.completed: https://developer.fastspring.com/reference/ordercompleted

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/ShareFileService.java
Dima-1: https://github.com/osmandapp/OsmAnd-tools/pull/1717/commits/19759c51471de4be2b76ab66e2d6815bbe93c3a5
agreed with the product owner

java-tools/OsmAndServer/src/main/java/net/osmand/server/utils/FileSizeFormatter.java
```diff
@@ -0,0 +1,12 @@
+package net.osmand.server.utils;
+
+import java.util.Locale;
+
+public class FileSizeFormatter {
```
alisa911: Locale.US is fine for the admin operation, but the share email is localized; pass the email's locale here (de/ru/fr expect a comma).

java-tools/OsmAndServer/src/main/java/net/osmand/server/utils/FileSizeFormatter.java
```diff
@@ -0,0 +1,12 @@
...
+
+	public static String format(long bytes) {
```
alisa911: No GB branch: a 1.5 GB file renders as "1536.0 MB".

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -1,20 +1,78 @@
...
 import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 @Service
 public class EmailSenderService {
+
+	private static final String SHARE_LINK_PREFIX = "https://osmand.net/map/share/join/";
```
alisa911: Every other email link is built in the template from @SITE_URL@ (LOGIN_URL, CLOUD_URL, ACTIVATE_URL); build FILE_URL the same way in cloud/share/approved/en.html and pass only FILE_UUID from here, so the host lives in one place.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -1,20 +1,78 @@
...
+				public void afterCommit() {
+					send.run();
+				}
+			});
+		} else {
+			send.run();
+		}
+	}
+
+	// language of the user's most recently used device (the app sends it on device-register, the web on login);
```
alisa911: The method name already says this; drop the comment.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
alisa911: No production caller left after this PR; the method and the cloud/web template are dead.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +109,189 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
 	}
     
     public void sendOsmAndCloudRegistrationEmail(String email, String token, String lang, boolean newUser) {
```
alisa911: Only caller is UserdataController:227; pass CloudAccountAction.SETUP / LOGIN there and drop this boolean wrapper.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +109,189 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+				.set("FILE_META", htmlText(meta))
+				.set("FILE_URL", htmlText(url))
+				.to(email)
+				.send()
+				.isSuccess();
+		LOGGER.info("sendShareFileAccessEmail approved=" + approved + " to: " + shorten(email) + " (" + ok + ")");
+	}
+
+	static String htmlText(String s) {
+		return s == null ? "" : org.springframework.web.util.HtmlUtils.htmlEscape(s).replace("@", "&#64;");
```
alisa911: Import HtmlUtils and move TEMPLATE_TOKEN_START up to the other constants.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +109,189 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+			renewalDate = skuData != null ? skuData.getExpireDate(iap.purchaseTime) : null;
+		} else {
+			return;
+		}
+		String productShort = productName.startsWith("OsmAnd ") ? productName.substring("OsmAnd ".length()) : productName;
+		EmailSenderTemplate sender = new EmailSenderTemplate()
+				.load("cloud/purchase/receipt", lang)
+				.set("EMAIL", htmlText(email))
+				.set("ORDER_ID", htmlText(orderId))
+				.set("ORDER_DATE", htmlText(formatReceiptDate(orderDate, lang)))
```
alisa911: Receipt uses the FastSpring checkout language while every other email uses userLang(); pick one source for the same user.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +109,189 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+		boolean ok = sender.to(email).send().isSuccess();
+		LOGGER.info("sendPurchaseReceiptEmail order " + orderId + " to: " + shorten(email) + " (" + ok + ") [" + lang + "]");
+	}
+
+	public void sendPurchaseLinkedEmail(String email, String lang,
+			List<DeviceInAppPurchasesRepository.SupporterDeviceInAppPurchase> purchases,
+			List<DeviceSubscriptionsRepository.SupporterDeviceSubscription> subscriptions) {
+		String productName;
+		if (subscriptions != null && !subscriptions.isEmpty()) {
+			DeviceSubscriptionsRepository.SupporterDeviceSubscription sub = subscriptions.get(0);
```
alisa911: The relink lists are unfiltered (findAllByUserId, findByUserId), so get(0) may be an expired or invalid purchase; there is findByUserIdAndValidTrue for inapps, subscriptions need the same.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +109,189 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+			PurchasesDataLoader.Subscription skuData = purchasesDataLoader.getSubscriptions().get(sub.sku);
+			productName = skuData != null ? skuData.name() : sub.sku;
+		} else if (purchases != null && !purchases.isEmpty()) {
```
alisa911: One-time inapps go through LINKED_INTRO "…subscription is now linked" and LINKED_FEATURES "All Pro features"; the text is wrong for Maps+ / lifetime purchases.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +109,189 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+				.set("EMAIL", htmlText(email))
+				.set("PRODUCT_NAME", htmlText(productName))
+				.to(email)
+				.send()
+				.isSuccess();
+		LOGGER.info("sendPurchaseLinkedEmail to: " + shorten(email) + " (" + ok + ") [" + lang + "]");
+	}
+
+	private static String formatReceiptDate(Date date, String lang) {
+		Locale locale = lang == null || lang.isEmpty() ? Locale.ENGLISH : Locale.forLanguageTag(lang.replace('_', '-'));
```
alisa911: Duplicates safeLang's lang parsing; also formats in the server time zone, which can shift "Renews on" by a day for the user.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/EmailSenderService.java
```diff
@@ -51,15 +109,189 @@ public void sendOsmAndCloudWebEmail(String email, String token, String action, S
...
+			locale = Locale.ENGLISH;
+		}
+		return DateFormat.getDateInstance(DateFormat.MEDIUM, locale).format(date);
+	}
+
+	public void sendOsmAndCloudAccountEmail(String email, String token, String lang, CloudAccountAction action,
+			String newEmail) {
+		EmailSenderTemplate sender = new EmailSenderTemplate()
+				.load(action.template, lang)
+				.set("TOKEN", token == null ? "" : token);
```
alisa911: Only EMAIL_CHANGED has no code; for the other actions a null token should fail loudly, not render an empty code box.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
```diff
@@ -172,16 +195,51 @@ public EmailSenderTemplate send() {
...
 	public EmailSenderTemplate load(String template, @Nullable String langNullable) {
-		String lang = langNullable == null ? "en" : langNullable;
+		String lang = safeLang(langNullable);
 		include("defaults", lang, false); // settings (email-headers, vars, etc)
-		include("header", lang, false); // optional
-		include(template, lang, true); // template required
-		include("footer", lang, false); // optional
-		include("unsubscribe", lang, false); // optional
+
+		if (checkUseBase(template, lang)) {
```
alisa911: checkUseBase does not set USE_BASE, so base/base-locale/header are still parsed with the blind strip (line 486) and any [if mso] block in header.html is deleted. Call set(USE_BASE, "true") here before the includes.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
```diff
@@ -172,16 +195,51 @@ public EmailSenderTemplate send() {
...
+		}
+
+		if (!"true".equals(vars.get(TRANSACTIONAL))) {
+			include("unsubscribe", lang, false);
+		}
 		return this;
 	}
 
+	private boolean checkUseBase(String template, String lang) {
+		File file = findTemplateFile(template, lang);
```
alisa911: he template file is read twice per email (here and in include); read it once and decide from the parsed lines.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
```diff
@@ -255,7 +313,7 @@ public boolean hasEmptyToList() {
 	private String fill(String in) {
 		String filled = in;
 		if (filled != null) {
-			final int PASSES = 2; // enough
+			final int PASSES = 4; // deep enough for base-template vars (@FOOTER_CONTACT@ -> _T2 -> @SUPPORT_EMAIL@)
```
alisa911: Both CODE_BLOCK and RENEWAL_ROW chains are already 4 levels, exactly PASSES (the comment describes a 3-level one). Loop until a pass replaces nothing instead of a fixed count.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
```diff
@@ -295,14 +353,20 @@ private void validateLoadedTemplates() {
 	}
 
 	private final String HTML_COMMENT_MATCH = "(?s).*<!--.*?-->*.";
```
alisa911: Same fix as HTML_COMMENT_REPLACE: HTML_COMMENT_MATCH and the two Pattern.compile calls on lines 373/377 still compile per line.

java-tools/OsmAndServerUtilities/src/main/java/net/osmand/mailsender/EmailSenderTemplate.java
```diff
@@ -295,14 +353,20 @@ private void validateLoadedTemplates() {
...
-	private final String HTML_COMMENT_REPLACE = "(?s)<!--.*?-->"; // (?s) Pattern.DOTALL (multiline)
+	private static final Pattern HTML_COMMENT_REPLACE = Pattern.compile("(?s)<!--.*?-->"); // (?s) Pattern.DOTALL (multiline)
 	private final String HTML_NEWLINE_TO_BR = "HTML_NEWLINE_TO_BR"; // user-defined var from templates
+	private final String TRANSACTIONAL = "TRANSACTIONAL";
+	private final String USE_BASE = "USE_BASE";
+	private static final Pattern USE_BASE_FLAG = Pattern.compile("\\s*<!--\\s*Set\\s+USE_BASE\\s*=\\s*true\\s*-->\\s*");
```
alisa911: This regex and parseCommandArgumentsFromComment disagree (set, Set: are accepted by the parser only), so a template can end up with USE_BASE=true in vars but no base wrapping. Reuse the parser for the check.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/ShareFileService.java
```diff
@@ -258,11 +262,36 @@ public boolean updateRequests(Map<Integer, String> accessMap) {
...
+					notifyAccessRequest(access, true);
+				} else if (PermissionType.BLOCKED.name().equals(accessType)) {
+					notifyAccessRequest(access, false);
+				}
 			}
 		}
 		return true;
 	}
 
+	private void notifyAccessRequest(ShareFileRepository.ShareFilesAccess access, boolean approved) {
```
alisa911: This is still the sending logic in the share service: owner/userFile lookups, the approve/decline choice and the try/catch. Move it to EmailSenderService (e.g. sendShareFileAccessEmail(ShareFilesAccess access, boolean approved)) and leave a one-line call here.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/FastSpringController.java
```diff
@@ -431,6 +433,19 @@ private void setInitialSubscriptionDates(DeviceSubscriptionsRepository.Supporter
...
```
alisa911: The receipt orchestration (date fallback, try/catch) is still in the controller; pass the event and the lists to EmailSenderService and keep only the sendAfterCommit call here.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/FastSpringController.java
```diff
@@ -148,6 +148,8 @@ private ResponseEntity<String> handleOrderCompletedEvent(FastSpringWebhookReques
 
 			userSubService.verifyAndRefreshProOrderId(user);
 
+			emailSender.sendAfterCommit(() -> sendPurchaseReceipt(email, event, purchases, subscriptions));
```
alisa911: The gift email on line 155 is still sent inside the transaction while the receipt waits for commit; wrap both in sendAfterCommit, or neither.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/FastSpringController.java
```diff
@@ -431,6 +433,19 @@ private void setInitialSubscriptionDates(DeviceSubscriptionsRepository.Supporter
 		}
 	}
 
+	private void sendPurchaseReceipt(String email, FastSpringWebhookRequest.Event event,
+	                                 List<DeviceInAppPurchasesRepository.SupporterDeviceInAppPurchase> purchases,
+	                                 List<DeviceSubscriptionsRepository.SupporterDeviceSubscription> subscriptions) {
+		FastSpringWebhookRequest.Data data = event.data;
+		try {
+			Date orderDate = event.created != null ? new Date(event.created) : new Date();
```
alisa911: Duplicates line 113; with created == null the receipt date is taken after commit and differs from the stored purchaseTime. Compute the order date once and reuse it.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/FastSpringController.java
alisa911: The receipt now shows "Renews on" for every FastSpring subscription because autorenewing is hardcoded to true here; the issue's Renews/Expires rule needs the real flag from the order item.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -542,7 +544,9 @@ public ResponseEntity<String> webUserRegister(String email, String lang, boolean
             }
             pu.tokenTime = new Date();
             usersRepository.saveAndFlush(pu);
-            emailSender.sendOsmAndCloudWebEmail(pu.email, pu.token, "@ACTION_SETUP@", lang);
+            emailSender.sendOsmAndCloudAccountEmail(pu.email, pu.token, lang,
+                    isNew ? EmailSenderService.CloudAccountAction.SETUP
```
alisa911: isNew=false is only sent by the web "Resend code" button during signup (CreateAccount.jsx:270), so this now emails "Change your password" to a user who is still creating the account.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
alisa911: Two sources of truth for the TTL (this constant and CODE_TTL_MINUTES in base.html); set CODE_TTL_MINUTES from sendOsmAndCloudAccountEmail so the email can never drift from the server.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -1420,7 +1423,7 @@ public void getBackupFolder(HttpServletResponse response, CloudUserDevicesReposi
     public ResponseEntity<String> deleteAccount(String token, CloudUserDevicesRepository.CloudUserDevice dev, HttpServletRequest request) throws ServletException {
         CloudUsersRepository.CloudUser pu = usersRepository.findById(dev.userid);
         if (pu != null && pu.id == dev.userid) {
-            boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() > TimeUnit.MILLISECONDS.convert(24, TimeUnit.HOURS);
+            boolean tokenExpired = System.currentTimeMillis() - pu.tokenTime.getTime() >= CODE_EXPIRATION_TIME_MS;
```
alisa911: tokenTime is null after registerNewDevice, so a logged-in user hitting this without a prior send-code gets an NPE instead of "Token is not valid"; guard it like wearOutToken does (same on 1508 and 1525).

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -1477,15 +1480,17 @@ private boolean deleteAllFiles(CloudUserDevicesRepository.CloudUserDevice dev) {
     }
 
     @Transactional
-    public ResponseEntity<String> sendCode(String action, String lang, CloudUsersRepository.CloudUser pu) {
-        if (!("setup".equals(action) || "change".equals(action) || "delete".equals(action))) {
+    public ResponseEntity<String> sendCode(EmailSenderService.CloudAccountAction action, String lang,
+            CloudUsersRepository.CloudUser pu) {
+        if (action == null) {
             return ok();
```
alisa911: Unknown action now arrives as null from fromCodeRequest; answering 200 without sending anything hides the client bug, return badRequest.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -1477,15 +1480,17 @@ private boolean deleteAllFiles(CloudUserDevicesRepository.CloudUserDevice dev) {
...
         }
         if (pu == null) {
             return ResponseEntity.badRequest().body("Email is not registered");
         }
         String token = (new Random().nextInt(8999) + 1000) + "";
-        emailSender.sendOsmAndCloudWebEmail(pu.email, token, action, lang);
+        emailSender.sendOsmAndCloudAccountEmail(pu.email, token, lang, action,
```
alisa911: The code email is sent inside the transaction before the token is stored; a rollback leaves the user with a code that never existed. sendAfterCommit is right there, use it.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserdataService.java
```diff
@@ -1441,7 +1444,7 @@ public ResponseEntity<String> deleteAccount(String token, CloudUserDevicesReposi
                     return ResponseEntity.badRequest().body(String.format("Unable to delete user files (%s)", pu.email));
                 }
             }
-            return ResponseEntity.badRequest().body("Token is not valid or expired (24h), or password is not valid");
+            return ResponseEntity.badRequest().body("Token is not valid or expired (10 min), or password is not valid");
```
alisa911: Four copies of "(10 min)" next to CODE_EXPIRATION_TIME_MS; derive the text from the constant.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -745,7 +745,13 @@ public ResponseEntity<String> sendCodeToNewEmail(@RequestParam String action, @R
 			usersRepository.saveAndFlush(pu);
```
alisa911: This endpoint only makes sense for change; setup/delete would email a setup or deletion code to the new address. Reject anything else before the temp user is created on line 741.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -745,7 +745,13 @@ public ResponseEntity<String> sendCodeToNewEmail(@RequestParam String action, @R
...
 
 			// send code to new email
-			return userdataService.sendCode(action, lang, pu);
+			EmailSenderService.CloudAccountAction emailAction = EmailSenderService.CloudAccountAction.fromCodeRequest(action, true);
+			ResponseEntity<String> sent = userdataService.sendCode(emailAction, lang, pu);
```
alisa911: sendCode answers 200 even when the email itself failed, so the old address can get the "being changed" notice while the new one got nothing. Gate this on the actual send result.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/user/MapApiController.java
```diff
@@ -713,7 +713,7 @@ public ResponseEntity<String> sendCode(@RequestParam String action, @RequestPara
 		if (pu == null) {
 			return ResponseEntity.badRequest().body("User not found");
 		}
-		return userdataService.sendCode(action, lang, pu);
+		return userdataService.sendCode(EmailSenderService.CloudAccountAction.fromCodeRequest(action, false), lang, pu);
```
alisa911: Static-import CloudAccountAction.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -223,7 +224,7 @@ public ResponseEntity<String> userRegister(@RequestParam String email,
 			// see comment on constant
 			pu.token = (new Random().nextInt(8999) + 1000) + "";
 			// TODO iOS: add lang in OARegisterUserCommand.m before sendRequestWithUrl params[@"lang"] = ...
-			emailSender.sendOsmAndCloudRegistrationEmail(pu.email, pu.token, lang, true);
+			emailSender.sendOsmAndCloudRegistrationEmail(pu.email, pu.token, lang, newUser);
```
alisa911: Sent inside the transaction before the user is saved and the order/supporter linking runs; a rollback leaves the user with a code for an account that does not exist. Use sendAfterCommit.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -199,6 +199,7 @@ public ResponseEntity<String> userRegister(@RequestParam String email,
 		// allow to register only with small case
 		email = email.toLowerCase().trim();
 		CloudUser pu = usersRepository.findByEmailIgnoreCase(email);
+		boolean newUser = pu == null;
```
alisa911: Template choice keys on pu == null while the client sends login; a web-created account with no devices gets "New device sign-in" on its first app sign-in. If that is intended, fine, but !Boolean.TRUE.equals(login) would match the client's intent.

java-tools/OsmAndServer/src/main/java/net/osmand/server/controllers/pub/UserdataController.java
```diff
@@ -479,7 +480,8 @@ public ResponseEntity<String> sendCode(@RequestBody EmailSenderInfo data,
 		if (pu == null) {
 			return ResponseEntity.badRequest().body("User not found");
 		}
-		return userdataService.sendCode(data.action, data.lang, pu);
+		return userdataService.sendCode(EmailSenderService.CloudAccountAction.fromCodeRequest(data.action, false),
```
alisa911: Static-import CloudAccountAction; unknown data.action still answers 200 with no email.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserSubscriptionService.java
```diff
@@ -486,6 +489,27 @@ public void relinkPurchasesToNewAccount(CloudUsersRepository.CloudUser previousU
 		previousUser.orderid = null;
 		usersRepository.saveAndFlush(previousUser);
 		LOG.info("Cleared orderId for previous user " + previousUser.id);
+
+		notifyPurchaseLinked(newUserId, iapList, subscriptionList);
```
alisa911: This is the only sender of cloud/purchase/linked; the first-time link in updateSubscriptionUserId / updateInAppPurchaseUserId still sends nothing.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserSubscriptionService.java
```diff
@@ -486,6 +489,27 @@ public void relinkPurchasesToNewAccount(CloudUsersRepository.CloudUser previousU
...
+	}
```
alisa911: Unfiltered lists (findAllByUserId, findByUserId) go to the email, which names get(0); pass only valid purchases (there is findByUserIdAndValidTrue for inapps, subscriptions need the same).

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserSubscriptionService.java
```diff
@@ -486,6 +489,27 @@ public void relinkPurchasesToNewAccount(CloudUsersRepository.CloudUser previousU
...
+
+	// Send only once the relink is committed, so a rollback cannot leave the user with an email about purchases
+	// that stayed on the previous account. A mail failure must not fail the relink itself.
```
alisa911: The code already says this (sendAfterCommit, try/catch); drop the comment.

java-tools/OsmAndServer/src/main/java/net/osmand/server/api/services/UserSubscriptionService.java
```diff
@@ -89,6 +89,9 @@ public class UserSubscriptionService {
 	@Autowired
 	private PromoService promoService;
 
+	@Autowired
```
alisa911: private, like the other autowired fields here.

java-tools/OsmAndServer/src/test/java/net/osmand/server/api/services/EmailSenderServiceEscapeTest.java
```diff
@@ -0,0 +1,99 @@
...
+	public void bulletproofButtonMarkersSurviveIntactWithNoHeaderPollution_whenUseBase() throws Exception {
+		EmailSenderTemplate t = new EmailSenderTemplate()
+				.set("USE_BASE", "true")
+				.template(MSO_SNIPPET)
+				.from("a@a.com").name("A").to("b@b.com");
+		String body = t.toString();
+
+		assertTrue("[if mso] opener must survive", body.contains("<!--[if mso]>"));
+		assertTrue("VML content must survive", body.contains("<v:roundrect href=\"#\">VML button</v:roundrect>"));
+		assertTrue("closing endif must survive", body.contains("<![endif]-->"));
```
alisa911: These two tests exercise EmailSenderTemplate, not EmailSenderService; move them to OsmAndServerUtilities/src/test/java/net/osmand/mailsender/ next to the class.

java-tools/OsmAndServer/src/test/java/net/osmand/server/api/services/EmailSenderServiceEscapeTest.java
```diff
@@ -0,0 +1,99 @@
...
+		assertTrue("[if !mso] opener must survive", body.contains("<!--[if !mso]><!-->"));
+		assertTrue("real anchor fallback must survive", body.contains("<a href=\"#\">Real button</a>"));
```
alisa911: Setting USE_BASE via set() bypasses load()/checkUseBase(), so the real include order is untested; a load() from a temp templates dir with an [if mso] block in header.html would expose that header is still blind-stripped.

java-tools/OsmAndServer/src/test/java/net/osmand/server/api/services/EmailSenderServiceEscapeTest.java
```diff
@@ -0,0 +1,99 @@
...
+		assertTrue("the [if !mso] opener must NOT survive outside USE_BASE - old blind-strip behavior",
+				!body.contains("<!--[if !mso]><!-->"));
+		assertTrue("the real anchor itself is plain HTML (no <!-- around it) and always survives",
+				body.contains("<a href=\"#\">Real button</a>"));
+
+		assertEquals("mso markers must never leak into email headers, even outside USE_BASE",
+				0, headersOf(t).size());
+	}
+
+	private static HashMap<String, String> headersOf(EmailSenderTemplate t) throws Exception {
```
alisa911: Reflection on a private field; a package-private headers() accessor on EmailSenderTemplate is cheaper than this.

alisa911 (changes requested):

PR #1722, opened 2026-09-16
Depth test tile styles: old vs new depth OBFs (test only)

vshcherb:
Test only, **not for merge** (OsmAnd-Issues#3341). Lets the test server show old and new depth maps side by side.

- Styles `depth-old-{marine,default,nautical}`: published depth OBFs (`download.osmand.net`, `depth=yes`) with the rendering styles from OsmAnd-resources `cf87cb06` (before #1478).
- Styles `depth-new-*`: OBFs from `builder.osmand.net/depth-data/build/` with the styles from OsmAnd-resources `master`.
- Depth contours are always on for these styles, tiles are `Cache-Control: no-store`.
- Everything goes to `java.io.tmpdir/osmand-depth-test` (~12 GB). The download starts on the first depth tile; until it finishes the tile answers 503 with the progress.
- The native library has one set of files: a depth style closes the other set (and `*.depth.obf` in the OBF folder) before rendering, so ordinary styles show whichever set was used last.
- `POST /tile/depth-test/clear-cache` drops the depth tiles (memory and disk) and re-downloads new OBFs whose size on the builder changed. Button: osmandapp/web#2037.

Checked locally: old → new → old on one server renders the right set, the base depth OBF is closed, clear-cache removes the tile folders and starts the refresh; `hd` tiles unchanged.

## AI disclaimer

Prompts used (summarised):
1. Draft PR in tools only, not to merge: test old vs new depth OBFs on the test server like the local two-server setup; download into temp from builder depth-data/build for new and the published maps for old.
2. A clear cache button under Rendering props.

Decided by the agent, not requested explicitly: one server with a native file-set swap instead of two servers; styles fetched from GitHub raw at a fixed commit; 503 while downloading; refresh by Content-Length on clear-cache; no-store cache header; forced depthContours; the button in a separate web draft PR.


vshcherb:
Replaced by #1723 (styles and maps built into the branch). (gen by AI)

PR #1725, opened 2026-09-17
Add registerNew/deleteFiles flags and id lists to notify-inactive-users

alisa911:
notify-inactive-users no longer deletes files on its own.

- `deleteFiles` (default false): delete files of users notified more than `graceDays` ago, prod run only.
- `registerNew` (default false): store new inactive users as notified, prod run only. Other runs only report them, so the list built on the first mail never grows.
- Result is comma-separated id lists per category: `new*`, `stillInactive*`, `reactivated*`, `eligibleForDelete*`, `deleted*`, each with a `…Count`.

Doc: osmandapp/OsmAnd-Issues#3342


alisa911:

PR #1728, opened 2026-09-19
Leave out an HH run the maps have no hub graph for

alex-dev-neo:
Build [#376](https://test.osmand.net) of `Maps_RandomRouteTester` went red on the pedestrian maps, and #1727 is why:

```
1:brp-java   cost=8078.93 dist=8735.00 segments=41767
1:brp-cpp    cost ok dist ok
1:brp-shared cost ok dist ok
1:hh-shared  cost zero dist zero segments=0     <- red
```

The pedestrian profile has no HH sections in the maps, so the planner raises `FAILED_NO_HH_ROUTING_DATA` and comes back with nothing. The tester counted that as a route of zero cost and zero distance - a failure. The job keeps the java and C++ HH runs away from pedestrian with `--avoid-hh-java --avoid-hh-cpp`; the shared runs it knows nothing about, so they ran and turned the build red.

An HH run over maps without a hub graph has no route to compare, so it is now left out of the results, with a line saying why:

```
hh-java: no HH routing data for pedestrian, skipped
hh-shared: no HH routing data for pedestrian, skipped
1:brp-java   cost=9120.05 dist=7939.08 segments=5860
1:brp-shared cost ok dist ok
```

It applies to the java and C++ planners as well, which makes those two `--avoid` options in the pedestrian line of the job unnecessary - they can stay, they only save the time of a search that finds nothing.

Checked on `Germany_hessen_kassel_europe.obf`: pedestrian leaves both HH runs out, car keeps all four and they agree to the digit.


alex-dev-neo:
Wrong fix: skipping the run hides the case it should catch. A car map built without its HH section would report FAILED_NO_HH_ROUTING_DATA just the same, and with this the run would be left out quietly and the job would stay green - today it goes red, which is what a map check wants. The job's pedestrian line takes --avoid-hh-shared beside the two it already has.

PR #1729, opened 2026-09-21
Keep a renewing subscription valid while the store retries payment

alisa911:
`UpdateSubscription` marked a subscription `valid=false, kind=expired` as soon as the expiry date reported by the store was older than 15 days, regardless of `autoRenewing`. Google keeps a subscription in account hold for up to 60 days and Apple retries billing for up to 60 days while the expiry date stays in the past; once payment succeeds the store renews the same purchase. By then our record was already `valid=false` and out of the job's selection (`valid is null or valid=true`), so the renewal never reached the database: the app showed the subscription as active, Cloud and the site reported it as expired.

Tickets #92321, #92185, #90887, #92275.

Now a subscription with `autoRenewing=true` is kept valid for up to 60 days after its expiry date. Access is unaffected during that time: `isProSubscriptionValid` still requires `expiretime > now`. The record stays in the selection and is re-checked every 12 hours, so the renewal is picked up as soon as the store reports it. Subscriptions with `autoRenewing=false` or `null` keep the previous 15-day rule.

Records that are already `valid=false` are not touched by this change; a one-off run with the existing `-revalidateinvalid` flag (or `-orderid=` for the tickets above) restores those the store still renews.

Docs: [Google Play subscription lifecycle](https://developer.android.com/google/play/billing/lifecycle/subscriptions), [Apple: reducing involuntary subscriber churn](https://developer.apple.com/documentation/storekit/reducing-involuntary-subscriber-churn).

## AI disclaimer

Written with Claude Code (Claude Fable 5.1), driven by [@alisa911](https://github.com/alisa911). Summary of the requests it worked from, in order:

1. Check whether the statement "a subscription the store keeps renewing is marked valid=false forever" is still true in the code. The agent traced every path that writes or restores `valid` (the job's selection and `updateSubscriptionDb`, `UserSubscriptionService` revalidation, `/subscription/purchased`, the FastSpring hooks) and reported which parts of the statement hold and which do not.
2. Explain the mechanism on four tickets with real rows. For #92185 (`paymentstate=0, autorenewing=true, checktime = expiretime + 15 days`) the agent identified the 15-day rule in `updateSubscriptionDb` ignoring `autoRenewing`/`paymentState` as the cause; the author rejected "temporary store failure" as a cause after the agent showed it does not lead to `valid=false` in the code.
3. Explain the same for Apple and FastSpring against their documentation. Apple: `is_in_billing_retry_period` and `grace_period_expires_date_ms` are not read, expiry is not extended during grace period, billing retry lasts up to 60 days. FastSpring: retries last days, and the `subscription.charge.completed` hook already restores the record, so the issue is limited to Google and Apple.
4. Propose a fix, then commit it with one-sentence comments linking the Google and Apple documentation. The author asked to undo the first commit to review the diff, confirmed the change is sufficient for the cause, and asked to commit, push and open the PR.

No tests exist for `updateSubscriptionDb`; the module was compiled by the agent. The one-off repair of existing `valid=false` rows is a server-side run, not part of this change. No commits or pushes were made without explicit permission.


alex-dev-neo:
**Transient store errors still expire a renewing subscription at day 15**

The new 60-day window only covers the success path. The four store-error handlers still apply the flat 15-day rule and end in `valid = false`:

| handler | line |
|---|---|
| `processHuaweiSubscription` | `UpdateSubscription.java:491` |
| `processAmazonSubscription` | `:559` |
| `processAndroidSubscription` | `:626` |
| `processFastSpringSubscription` | `:687` |

```java
if (expireTime != null && currentTime - expireTime.getTime() > MAX_WAITING_TIME_TO_EXPIRE) {
    kind = EXPIRED_STATE;
}
...
deleteSubscription(orderId, sku, currentTime, reason, kind);
// delQuery: UPDATE supporters_device_sub SET valid = false, kind = ?, checktime = ?
```

Scenario: a Google subscription in account hold, `autoRenewing=true`, expired 20 days ago. With this PR it stays `valid=true` and is re-checked every 12 h. On one of those runs `purchases.subscriptions().get()` throws `IOException` (quota, 5xx, network blip). The catch takes the first branch, `kind=expired`, `deleteSubscription` writes `valid=false`, and the row leaves the job's selection `(valid is null or valid=true)` for good — so the renewal on day 25 never reaches the database. That is the same end state this PR fixes, reached through a different path.

Before this change the row was already `valid=false` at day 15, so the error path never mattered. Now it is polled twice a day for 60 days instead of 15, so there are ~4x more runs in which one transient failure is terminal.

Suggestion: give the error handlers the same rule as the success path — extract

```java
static boolean isExpired(long expiryMillis, Boolean autoRenewing, long now)
```

next to the constants and call it from `updateSubscriptionDb` and from all four catch blocks. This needs `autorenewing` added to `selQuery` (line 163) and passed into the `process*` methods — the column is not selected today, so the last known value is not available in the catch blocks.

alex-dev-neo:
**The 60-day window is undone for FastSpring by `UserSubscriptionService`**

`updateSubscriptionDb` is shared by all five stores, so the new rule also applies to FastSpring rows. FastSpring has a second writer that uses a different rule, and it wins.

`isProSubscriptionValid` revalidates whenever `expiretime` is in the past (`UserSubscriptionService.java:130`), and for FastSpring that reaches `applyFastSpringSubscription`, whose normal branch is:

```java
// UserSubscriptionService.java:398
s.valid = now < s.expiretime.getTime();
// revalidateFastSpringSubscription:365-366 then calls subscriptionsRepo.save(s)
```

So for a FastSpring subscription in dunning (`active=true`, `autoRenew=true`, `state=overdue`, `getExpiryTime()` returning `next` in the past) the job writes `valid=true`, and the next time any endpoint reaches `checkOrderIdPro` for that order the web path writes `valid=false` and saves. The row then no longer matches `(valid is null or valid=true)`, so the job never touches it again: the web path wins permanently and the FastSpring half of this change is silently reverted.

The other stores are safe only by accident — `revalidateGoogleSubscription` / `Huawei` / `Amazon` / `iOS` save **only** when the store reports a later expiry (`s.expiretime.getTime() < subscription.getExpiryTimeMillis()`, lines 258, 284, 308, 336), which is false during account hold, so they leave `valid` alone. `applyFastSpringSubscription` has no such guard.

Two ways out:
- keep the 15-day rule for FastSpring — the PR description already argues the issue is limited to Google and Apple, since `subscription.charge.completed` restores the record; or
- apply the same rule in `applyFastSpringSubscription` so both writers agree.

alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3352

PR #1730, opened 2026-09-21
TopPhotos: admin pages and ClickHouse queries of the top places photo scoring moved into OsmAndServer

alisa911:
Moves the internal "top places photos" admin tool (the Express server on maptile.osmand.net:4000) into OsmAndServer, so the port can be closed.

`TopPhotosController` under `/admin/top-photos` serves the five pages (`/`, `/per-place`, `/filter`, `/stat`, `/diff`) and the 15 JSON endpoints under `/admin/top-photos/api/*`, reading the wiki ClickHouse through `wikiJdbcTemplate`. `/admin/**` is already `ROLE_ADMIN` only. The Thymeleaf templates and their js/css come with osmandapp/web-server-config (`templates/admin/top-photos`, `js/top-photos`, `css/top-photos.css`); the header link in osmandapp/web opens the pages.

Queries are copied from `top-photos-backend.js` / `top-places-backend.js` one to one, with these differences:

- every user value is a `?` parameter, sort columns are whitelisted (the node code interpolated `tagFilter`, `quidFilter`, `sortCol`, `runSelect` and the free-text search into the SQL)
- `filter-runs` filtered by `proc_id`, which `top_images_run` does not have, so the query failed whenever a place was selected; it now uses `wikidata_id`
- `edit-place` stores the admin's login in `userid` instead of the request IP
- the SQL console (`top-photos-sql.html`, `execute-sql`) is not ported, it is not wanted any more
- `/photo/api/runs` is not ported: it called a function that was never imported and always returned 500, no page uses it

## AI disclaimer

Written with Claude Code (Claude Fable 5.1), driven by [@alisa911](https://github.com/alisa911). Summary of the requests it worked from, in order:

1. Find where https://maptile.osmand.net:4000/top-photos-per-place comes from. Traced to the Express server `web-server-config/servers/wikidata/top-places/server.js`, its pages and the ClickHouse tables filled by the `tools/python/top-photos` jobs.
2. Move the page into the web prototypes with a header link like the heatmap. The agent pointed out the page needs the `/photo/api` endpoints that only exist on port 4000 and proposed either calling them cross-origin or porting them to OsmAndServer.
3. Port the two queries to OsmAndServer, the port should be closed later; check how ClickHouse is used on test. Found the `wikidatasource` config in `application.yml`, `wikiJdbcTemplate` used by `WikiService`, and verified `/search/get-wiki-images` on test answers from ClickHouse.
4. List everything on port 4000, explain what the API and Express are and whether the server is needed. Six pages and 17 endpoints listed, including the unauthenticated `execute-sql`.
5. Move everything server side into one controller without breaking anything, pages under `/admin/top-photos` as the equivalent of the 4000 root. Found the admin templates live in `web-server-config/templates/admin` and `/admin/**` is already restricted to `ROLE_ADMIN`; wrote `TopPhotosController`, moved the pages, js and css, and updated the URLs in `SCHEMA.md` and the python README. The agent flagged the `filter-runs` bug, the `userid` change and the dropped `/photo/api/runs`.
6. Add an admin-only header button with a dropdown of the pages in the web (separate PR in osmandapp/web).
7. Why is the header URL hardcoded to test, and the assets do not load locally. Replaced with `window.location.origin` and proxied `/admin/`, `/js/` and `/css/` in the web `setupProxy.js`.
8. Create a server branch, commit, push and open this PR.
9. Drop the SQL console, it is not wanted any more. Removed the `/sql` page and the `execute-sql` endpoint.

The Java part compiles (`:OsmAndServer:compileJava`); the pages were opened by the author against a local OsmAndServer with the wiki ClickHouse, the queries were not run by the agent. The agent made the commit and opened the PR on request.


alisa911:
https://github.com/osmandapp/web/issues/2014

PR #1731, opened 2026-09-21
Index: write the abbreviated spelling of a name into the name index

vshcherb:
Draft — the map-side half of a pair; the engine-side half is osmandapp/OsmAnd#26165. Either one alone fixes the same defect, and they differ in who gets the fix.

### The defect

`2 Trinity Pl, Hackensack` finds nothing: the map spells the street `Trinity Place`, the query writes `Pl`. Expanding it at search time (#26165) works only for what the installed app shipped with — a new abbreviation reaches users with the next app release, and never reaches the ones who do not update.

### This change

The name index already stores a second spelling of a name: `addUngluedToNameIndex` indexes `L'abordage` as `abordage` as well. This adds the same for abbreviations — `Trinity Place` is indexed as `Trinity pl` too — so the query matches the name the map carries, with no engine change at all. A map built after this makes the abbreviation searchable in an app that is already installed.

`addUngluedToNameIndex` and `addAbbreviatedToNameIndex` now share `addExtraSpellingToNameIndex`; both are called for addresses (`NameIndexCreator.putAddrNamedMapObject`) and for POIs (`IndexPoiCreator`).

Only abbreviations the runtime table cannot expand are written:

```java
private static final Map<String, String> EXTRA_ABBREVIATIONS = Map.of(
        "place", "pl", "parkway", "pkwy", "mount", "mt");
```

That limit is measured, not cautious. Writing every abbreviation the runtime already knows (`av`, `st`, …) makes each of them a full name word of nearly every street, and that changes ranking: with the broad list, `SpatialSearchPipelineTest` fails 2 of 85 — `usa_new_york` puts `Route 4 East At Forest Avenue, Paramus` above `South 4th Avenue, Mount Vernon` for `4 av`, and `translation` shifts an internal atom id in the golden. With the narrow list, nothing changes.

### Regressions

OBFs regenerated by the patched generator (the test hashes the generator classes, so the cache is invalidated automatically), stock search engine:

- `SpatialSearchPipelineTest` 85/85, `SpatialSearchScoreTest` 85/85, `SpatialSearchCombinationsTest` 85/85 — the score goldens compare 1,645 expected rows over 272 phrases, none changes.
- The same three suites are also 85/85 with #26165 applied on top, so the two PRs can land together.
- `UnglueNameTest` 11/11 after the shared-helper refactor; `AbbreviateNameTest` added (4 cases).

### What is not measured

The end-to-end effect on a real map: it needs a full New Jersey rebuild, which I did not run. On a hand-built OBF the effect on the name index is visible directly with a stock engine — the query word that was left unmatched before is now matched by the object:

```
stock generator : ["2 trinity"     BUILDING '2 Trinity Place']   <- "pl" unmatched
this generator  : ["2 trinity pl"  BUILDING '2 Trinity Place']
stock generator : ["mount vernon"     STREET 'Mount Vernon Parkway']
this generator  : ["mount vernon pkwy" STREET 'Mount Vernon Parkway']
```

Index growth is one extra token per name that contains one of the three words.

## AI disclaimer

Prompts used (summarised):
1. Check what the bug is, from a report on NY/NJ spatial-search defects (16 classes).
2. Dig for the root cause in the code for all of them.
3. What is the minimal fix in search, and what is the minimal fix in the index engine, as with unglue?
4. Create two draft PRs to study — one on the search engine, one on the index.
5. Check how critically they affect the current golden state of the score tests.

Decided by the agent, not requested explicitly: reusing the unglue mechanism rather than adding a new index section; restricting the list to abbreviations the runtime cannot expand after the broad list broke two goldens; the three entries themselves; the new unit test.


vshcherb:
Closing: redundant. The same defect is fixed on the engine side in osmandapp/OsmAnd#26165, and one side is enough — this one writes an extra token into the index for every name containing Place/Parkway/Mount, for no gain once the runtime table knows them. (gen by AI)

